Issue #9, opened 2017-04-11
Common extensions


kivikakk:
* [x] Task lists (#8)
* [x] ~Footnotes (#11)~ I think it's not appropriate.
* [x] Superscript (#10)
* [x] ~No intra-emphasis~ CommonMark does this better by default.

PR #11, opened 2017-04-11
Footnotes

kivikakk:
See #9.


kivikakk:
This is a significant extension and adds a lot of HTML with assumed IDs, classes, etc.; it doesn't feel appropriate for this layer.

PR #12, opened 2017-04-12
Make clippy happy

killercup:
This resolves all the important stuff clippy complained about. A few minor lint suggestions are not addressed (combining match arms with the same right side for examples). A lot of diff churn because I ran `s/'(.+?) as u8/b'$1'` without making a separate commit. Sorry.


kivikakk:
Great, thank you so much! I had no idea clippy was a thing, and the `b'x'` syntax was sorely needed. Thank you! :heart:

PR #13, opened 2017-04-12
Add some default capacities to strings

killercup:
I have not seen an easy way to actually benchmark this implementation, but while writing #12, I noticed some easy ways to prevent some reallocations.

### DIFF
--- a/src/entity.rs
+++ b/src/entity.rs
@@ -87,9 +87,9 @@ fn lookup(text: &str) -> Option<&'static str> {
 }
 
 pub fn unescape_html(src: &str) -> String {
-    let mut i = 0;
-    let mut v = String::new();
     let size = src.len();
+    let mut i = 0;
+    let mut v = String::with_capacity(size);
 
     while i < size {
         let org = i;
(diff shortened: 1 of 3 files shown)


kivikakk:
Hey, thank you so much for the PR! Looks good :+1:

For benchmarking, I've been using this invocation:

```
(cd vendor/cmark-gfm/; make bench PROG=../../target/release/comrak)
```

(Omitting `PROG` benches the reference C implementation if it's built.)

It's not very precise or well-rounded (e.g. it doesn't exercise any extensions), but imo it's fine for a rough approximation. I don't get a noticeable improvement benching this vs. current master, but that doesn't mean it's not an improvement! Thank you!

killercup:
Ah, thanks for the info! I've just used the benchinput generated by the cmark-gfm makefile to profile comrak a bit (using `perf`). Looks like `comrak::parser::Parser::parse_inlines` and `comrak::html::HtmlFormatter::escape` are each responsible for about ~25% of the runtime, so it might be a nice idea to optimize those.

Btw, to get the absolute best performance you should probably try to add `[profile.release] lto = true` to `Cargo.toml` (enables link time optimizations and increases inlining threshold) and compile with `env RUSTFLAGS="-C target-cpu=native" cargo build --release`.

kivikakk:
Thank you so much for that info! Somehow I hadn't thought of using normal benchmarking tools like `perf` on Rust binaries. I'm looking into `parse_inlines` and `escape` now. I've added the LTO flag to `Cargo.toml` too :3

PR #14, opened 2017-04-15
Optimise

kivikakk:
With suggestions from @killercup in https://github.com/kivikakk/comrak/pull/13#issuecomment-293943313; preliminary benchmarks look like we're at a reduction to 50% of the runtime in this PR.


killercup:
Sweet! 🎉

PR #20, opened 2017-05-25
Add static lifetime

gjtorikian:
On rustc 1.16.0, a simple `cargo build` errored with:

```
   Compiling comrak v0.1.4 (file:///Users/gjtorikian/Developer/comrak)
error: this needs a `'static` lifetime or the `static_in_const` feature, see #35897
 --> src/entity_data.rs:5:24
  |
5 | pub static ENTITIES: [(&str, &str); 2125] =
  |                        ^

error: this needs a `'static` lifetime or the `static_in_const` feature, see #35897
 --> src/entity_data.rs:5:30
  |
5 | pub static ENTITIES: [(&str, &str); 2125] =
  |                              ^

error: aborting due to 2 previous errors

error: Could not compile `comrak`.

To learn more, run the command again with --verbose.
```

So, I'm just doing what the compiler is suggesting.


kivikakk:
Ah — I'm following the latest stable here. (Static lifetimes became optional for global statics/consts in 1.17.) Do you think that's a reasonable thing to do?

gjtorikian:
> Do you think that's a reasonable thing to do?

It's your call! I didn't know they changed recently.

kivikakk:
K! I think I'll keep it as is, if only because I think encouraging targeting the latest stable is a Good Thing. Thank you for the PRs! :heart: :3

PR #21, opened 2017-05-26
Use the right dir

gjtorikian:
I stole your script for a project I'm working on and am fixing the comment from whichever project you stole it from. 😉


gjtorikian:
Actually, while looking for something unrelated, I came across https://github.com/p-jackson/entities, so I could also upgrade this project to use that, if it makes sense.

kivikakk:
Thank you! :D I stole this script directly from [`jgm/cmark`](https://github.com/jgm/cmark). I would like to use something more robust (like the linked repo), so long as it behaved acceptably similar to the upstream such that the specs still pass :3

PR #22, opened 2017-05-29
Use entities crate

gjtorikian:
As discussed in https://github.com/kivikakk/comrak/pull/21, this swaps out the current "array lookup" logic with a crate that implements similar functionality for figuring out entity characters.

This is my first "major" contribution to a Rust project so if I did something wrong, let me know! 😅


src/entity.rs
```diff
@@ -58,32 +62,18 @@ pub fn unescape(text: &str) -> Option<(String, usize)> {
...
+fn lookup(text: &str) -> Option<&str> {
+    let entity_str = format!("&{};", text);
+
+    let entity = ENTITIES
+        .iter()
+        .find(|e| e.entity == entity_str);
+
+    match entity {
+        Some(e) => {
+            Some(e.characters)
```
kivikakk: Minor nit: this `match` block can be replaced with `entity.map(|e| e.characters)`.

src/entity.rs
```diff
@@ -58,32 +62,18 @@ pub fn unescape(text: &str) -> Option<(String, usize)> {
...
-            low = i + 1;
-            i = j;
-        } else {
-            return None;
+fn lookup(text: &str) -> Option<&str> {
+    let entity_str = format!("&{};", text);
+
+    let entity = ENTITIES
+        .iter()
+        .find(|e| e.entity == entity_str);
```
kivikakk: This lookup is O(_n_) in the size of the `ENTITIES` array. The old code did a binary search, O(log _n_), and I think we want to try to preserve that performance characteristic — otherwise a page full of `&zwnj;` will result in traversing all 2,200 members of the array for each lookup.

One way to do that is to do as  [`entities`' README](https://github.com/p-jackson/entities#readme) suggests and create a `HashMap` from the `ENTITIES` data, e.g. in a `lazy_static!` section. This would be a good first-level speedup. Possibly even faster would be to construct our own sorted array from the `ENTITIES` data such that we could still do the binary search, but the jury's out onto whether that or `HashMap` (or `BTreeMap`) would actually be faster.

(One step further would be to create the entity data lookup array at _build time_ with a `build.rs`, which would get us back to where we started, albeit sans Python.)

src/entity.rs
```diff
@@ -1,7 +1,11 @@
 use ctype::isdigit;
-use entity_data;
 use std::char;
-use std::cmp::{min, Ordering};
+use std::cmp::min;
+
+use entities::ENTITIES;
+
+pub const ENTITY_MIN_LENGTH: usize = 2;
+pub const ENTITY_MAX_LENGTH: usize = 31;
```
kivikakk: Ideally we'd calculate these from `ENTITIES` directly, otherwise we might get surprised one day if there's a new release of the `entities` crate and they no longer hold.

PR #26, opened 2017-07-18
Fix quadratic behaviour during emphasis processing

### LINKED ISSUE #24: Quadratic blowup on pathological input.
This takes ~1s on my machine:

```sh
perl -e 'print "_a* " x 40000' | time comrak >/dev/null
```
This allows a denial of service in any application using Comrak to process untrusted input, and thus prevents it from being used for that purpose, forcing the use of a C library such as cmark or MD4C instead.

kivikakk:
Fixes #24. Uses `remstack.rs` provided by @charliesome.

Not quite ready to merge yet; the correctness isn't 100%.


kivikakk:
Next step: remove all messing around with indices entirely with opaque index/iterator type.

zeantsoi:
This is a test^(of) this

kivikakk:
@zeantsoi I can recommend using a test repo that doesn't already have users with notifications subscribed 🐰

kivikakk:
RemStack idea needs some development still, so for now I've rewritten this to use an Arena, keeping a lot closer to the C codebase.

PR #33, opened 2017-08-20
Avoid taking quadratic time

DemiMarie:
This avoids taking quadratic time by refusing to parse recursive
chevrons and limiting nested parens to 32.


kivikakk:
Thanks so much! 💕

PR #35, opened 2017-08-25
Fix for inter-unicode slices and indexing

SSJohns:
There was an error with slices in autolink where a unicode character would cause the parser to panic when it was a multi-character entity. For instance `[ww𨭎` covers 4 spaces and the line in autolink would try to slice into that. 

The change in inlines.rs is just a bounds check. All test still pass.


kivikakk:
Thank you! :heart:

Issue #25, opened 2017-07-11
Comrak is much slower than CMark


DemiMarie:
Running the spec tests with Comrak takes about 1.9 times as long as it does with CMark.

kivikakk:
This is correct! It's unclear how to improve it currently; pull requests and analyses welcome.

DemiMarie:
@kivikakk I created a flamegraph using `perf` and `flamegraph.pl`, and concluded that a massive 37.26% of the time is spent in the regex engine.  That can probably be sped up substantially (together with a massive decrease in code size) by hand-coding the scanners.

I suspect that eliminating internal copies can more than make up for the rest.

kivikakk:
@DemiMarie Oh, wonderful! Thank you so much; I haven't had the time recently to do this, so I appreciate your looking into it so much.

I'm away from the computer until next week, but when I get back I'll look into optimising the scanners per #16. After that, I'll see how much copy elimination I can do — I had a hack at this [using `tendril`](https://github.com/kivikakk/comrak/compare/tendril) a while back, so I'll have to revisit that work.

I also wonder how much overhead there is in simple indirection through `RefCell` for all AST nodes.

DemiMarie:
@kivikakk It might be possible to reduce even more overhead by not bothering with `tendril`’s reference-counting, and instead just storing indexes into the original string.  Rust’s lifetimes can guarantee that the string outlives the AST.

Another source of overhead is manipulating strings, rather than byte arrays.

SSJohns:
@DemiMarie Could that really b causing that much overhead? I started on a branch which tries to remove all strings and turn them into &str and use byte searching on them, but I wasn't sure how much it would actually improve performance.

DemiMarie:
I am not sure (I don’t know how to find it – I think it would show up as
uniformly slow code), but I know that pulldown-cmark and MD4C (the fastest
CommonMark implementation, by far) use it.  MD4C doesn’t even do UTF-8
validation – it promises that the output will be valid UTF-8 if the input
is, and that invalid UTF-8 will not cause undefined behavior, but in
general invalid UTF-8 input will result in invalid UTF-8 reaching the
rendering callbacks.

On Thu, Jul 20, 2017 at 2:18 PM, Shaquille Johnson <notifications@github.com

kivikakk:
[...]

PR #38, opened 2017-09-11
String->Vec<u8> optimisations

kivikakk:
Refs #25.

Replace `String`/`str` with `Vec<u8>`/`[u8]`. I'm not entirely pleased with the amount of `unsafe` blocks this adds where we reassert the input data is indeed UTF-8.

`make bench` comparison:

```
master: mean = 1.7500, median = 1.7500, stdev = 0.0694
opt:    mean = 1.6210, median = 1.6050, stdev = 0.0691
```

/cc @DemiMarie

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ entities = "1.0.0"
 unicode_categories = "0.1.1"
 clap = { version = "2.22.2", optional = true }
 clippy = { version = "~0.0.123", optional = true }
+twoway = "0.1.3"
 
 [features]
 default = ["clap"]
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ let arena = Arena::new();
 
 let root = parse_document(
     &arena,
-    b"This is my input.\n\n1. Also my input.\n2. Certainly my input.\n",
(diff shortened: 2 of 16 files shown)


DemiMarie:
What happens if we check for UTF-8 validity at the end?  Bulk validation
might be quite fast.

Also, I believe that there are faster UTF-8 checking algorithms than the
one in libstd.  The one I am thinking of is a table-based state machine,
and can validate UTF-8 branchlessly.

kivikakk:
We actually don't check for UTF-8 validity ourselves; that's why the unsafe blocks exist, because I'm just asserting they're safe (by using unchecked variants) rather than actually performing the check. It's just a bit ugly.

kivikakk:
(The input is a `str` and so is already guaranteed to be UTF-8 for us.)

DemiMarie:
Theoretically, if comrak had a bug, it could produce bogus output by
splitting along a code point.  Hence me asking about the cost of the
assertion.

kivikakk:
Makes sense! It'd be good to perform that more cheaply in `markdown_to_html`.

kivikakk:
Some experimentation replacing everything with `Cow<'a, [u8]>` in https://github.com/kivikakk/comrak/tree/cow.

PR #53, opened 2018-03-27
Fix a corner case in the ATX header parser

brson:
Hi @kivikakk.

Here's a fix for a panic in the parser. This is the patch I committed to Reddit's downstream fork, but I'm only just learning the codebase, and would love to have your feedback on the correct fix. The patch contains a massive comment indicative of my low understanding of the code in general, so I understand if you want that slimmed down.

Problem description from the commit follows.

Fix a corner case in the ATX header parser

The parser would previously crash in some cases of empty ATX headers with
trailing hashes. This case isn't captured by the CommonMark spec.

The bad cases look like this:

```
 ###    ####
```

(note the leading space here is not part of the bug - it's just so `git commit`
doesn't throw these lines away as comments)

In these cases the line parser advances `first_nonspace` all the way to the
second set of hashes, while the `chop_trailing_spaces` routine for stripping
the trailing hashes also trims leading whitespace, leading to a buffer
overrun as `add_text_to_container` thinks the content of the line
is only the leading "###" but that the text begins at the start
of the start of "####".

This hasn't been detected previously because the spec only contains
a test case for

```
 ### ###
```

and standard CommonMark always eats exactly one space after the initial ATX
hashes, leaving the line parser's `offset` equal to `first_nonspace`, and
correctly adding empty text to the header container.

The spec might want an additional test case containing multiple spaces in an
empty bookended ATX header.

cc @SSJohns

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -949,8 +949,25 @@ impl<'a, 'o> Parser<'a, 'o> {
                             }
                         };
                         let count = self.first_nonspace - self.offset;
-                        self.advance_offset(&line, count, false);
-                        self.add_line(container, &line);
+
+                        // In a rare case the above `chop` operation can leave
+                        // the line shorter than the recorded `first_nonspace`
+                        // This happens with ATX headers containing no header
+                        // text, multiple spaces and trailing hashes, e.g
+                        //
+                        // ###     ###
+                        //
+                        // In this case `first_nonspace` indexes into the second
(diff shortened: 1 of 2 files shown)


kivikakk:
This looks good to me; I think because we're actually shortening the string buffer with `truncate` we're hitting this panic; `cmark` just reduces the `len` of the buffer, but the underlying allocated memory remains and so it doesn't hit this.

Thank you!

PR #55, opened 2018-04-10
Fix the grammar for scanning table marker rows

brson:
Scan all the way to newline. This makes some edge cases
parse in accordance with upstream cmark.

Looks like this might have been a typo in the grammar.

I've confirmed this doesn't regress comrak against the cm spec (it has 5 existing failures), though I don't know where the gfm spec source is.

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -673,3 +673,21 @@ fn no_panic_on_empty_bookended_atx_headers() {
         "<h1></h1>\n"
     );
 }
+
+#[test]
+fn table_misparse_1() {
+    html_opts(
+        "a\n-b",
+        "<p>a\n-b</p>\n",
+        |opts| opts.ext_table = true,
+    );
+}
+
+#[test]
+fn table_misparse_2() {
+    html_opts(
+        "a\n-b\n-c",
+        "<p>a\n-b\n-c</p>\n",
+        |opts| opts.ext_table = true,
+    );
+}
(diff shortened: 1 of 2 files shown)


src/tests.rs
```diff
@@ -673,3 +673,21 @@ fn no_panic_on_empty_bookended_atx_headers() {
...
+}
+
+#[test]
+fn table_misparse_2() {
+    html_opts(
+        "a\n-b\n-c",
+        "<p>a\n-b\n-c</p>\n",
+        |opts| opts.ext_table = true,
+    );
+}
```
brson: This would previously parse as a 1-column table with header "a" and 1 row of content, "c", parsing "-b" as a valid table marker row.

kivikakk:
This is perfect, thank you! Indeed, this now matches https://github.com/github/cmark/blob/bf28ef6e7e949157ab85a5e4a3ea9107e17497b9/extensions/ext_scanners.re#L41 correctly.

The Travis CI build checks against the current version of the CM spec we're tracking, including the GFM specs; see https://travis-ci.org/kivikakk/comrak/jobs/364882300. I might need to pull a later version of upstream in to see what new failures are.

kivikakk:
> I might need to pull a later version of upstream in to see what new failures are.

I've pulled in the latest upstream and matched the `cmark` implementation changes in c7ac6ac.

PR #56, opened 2018-04-12
Thoroughly comment `process_emphasis` and `insert_emph`

brson:
Hi @kivikakk :)

Ok, consider this just a request for comments. I totally understand if you don't want this in your tree. No pressure at all. And preemptive apologies that this is going to occupy some of your time.

One of the things I've noted as I've been ramping up on comrak is that there aren't a lot of comments explaining the algorithm. As I was adding a new kind of emphasis to downstream snoomark, I found myself thoroughly reading the code and writing comments.

But before I committed I realized that all this noise would make the diff between comrak and snoomark a lot bigger, so I'm torn about what to do. Here I'm seeing how you feel about having this massive comment-bomb in-tree, but also looking for your feedback about whether I'm understanding things right.

And FWIW I read your awesome "inlines and new syntax" email to @SSJohns _after_ I did this - the comment on `odd_matches` is just copied directly from you because I couldn't figure it out :)

So a few things I've noted (in this patch), that I'm wondering if you can confirm:

`openers_bottom` is not initialized for `~` and `^`, but it seems like for maximum correctness it should be.

There _seems_ to be a bunch of dead code in `process_emphasis` for quotes because `handle_delim` doesn't push delimiters for them, and it _seems_ like both types of quotes might as well just not be processed as special chars at all.  Is that your understanding too? Is it in anticipation of future work (smart quotes), or just dead (can I remove it)?

And ... that's it.

Thanks for your time.

### DIFF
--- a/src/parser/inlines.rs
+++ b/src/parser/inlines.rs
@@ -158,8 +158,63 @@ impl<'a, 'r, 'o, 'd, 'i> Subject<'a, 'r, 'o, 'd, 'i> {
         }
     }
 
+    // After parsing a block (and sometimes during), this function traverses the
+    // stack of `Delimiters`, tokens ("*", "_", etc.) that may delimit regions
+    // of text for special rendering: emphasis, strong, superscript,
+    // spoilertext; looking for pairs of opening and closing delimiters,
+    // with the goal of placing the intervening nodes into new emphasis,
+    // etc AST nodes.
+    //
+    // The term stack here is a bit of a misnomer, as the `Delimiters` actually
+    // form a doubly-linked list. Items are pushed onto the stack during parsing,
+    // but during post-processing are removed from arbitrary locations.
+    //
+    // The `Delimiter` contains references AST `Text` nodes, which are also
+    // linked into the AST as siblings in the order they are parsed. This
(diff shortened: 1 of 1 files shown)


kivikakk:
Hey @brson, thanks so much for the PR!

I thought about it, and y'know what? I'm totally fine with having this in the tree. It seems likely that your fork of comrak is likely to be the largest actual production deployment, so keeping the diff smaller is :100: from my point of view.

The comments are all accurate! So thank you for that. Re: your notes:

* Yep, `openers_bottom` should indeed be initialised for the extension characters too.
* It is indeed dead code! I never got around to pulling smart quote handling in — there's a bit of an indication of this here:

  https://github.com/kivikakk/comrak/blob/40c4b306e8ccbd497677a7028c77c270e9ecad8d/src/parser/inlines.rs#L106-L108

  Ideally I'd actually implement that instead!

I'll merge this right away, but please feel free to ask more questions! (And I'll get back to your email c/o vmg during tomorrow working hours.)

brson:
Thanks @kivikakk ❤️

kivikakk:
I've cleared out these TODOs (including implementing smart characters). Thanks for the push!

PR #59, opened 2018-04-22
Make AstNode implement Debug in a useful way

brson:
This makes the Debug impl for arena_tree nodes do something useful, instead of ilooping on internal cycles.

I've used his several times for debugging.

cc  @kivikakk


brson:
👓 @SSJohns

kivikakk:
Thank you! :heart:

PR #61, opened 2018-04-26
Fix parsing of tables nested in other block elements

brson:
The row parser was not scanning from the correct offset.

### DIFF
--- a/src/parser/table.rs
+++ b/src/parser/table.rs
@@ -88,7 +88,7 @@ fn try_opening_row<'a, 'o>(
     if parser.blank {
         return None;
     }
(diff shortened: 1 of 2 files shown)


kivikakk:
I hope this is okay with you — I've added you as a collaborator and marked `master` as protected, so you can merge yourself after review.

kivikakk:
In fact, I've revised my decision: feel free to merge without a review first if you're confident. (And this is a pretty clear example of something to be confident about.) I'm a big fan of [Peter Hintjens' "optimistic merging" idea](http://hintjens.com/blog:106).

PR #63, opened 2018-05-03
Fix stack smashing by converting recursion to iteration

brson:
It's possible to feed comrak input that causes it to run off the end of the stack, e.g., a long series of ">>>>" is a nested blockquote and will recurse deeply.

This eliminates all the recursion I could find.

Note that the change to postprocess_text_nodes actually changes the order nodes are visited, and in a way that could affect correctness and cache behavior. I benchmarked the change and couldn't see any conclusive difference in perf.

### DIFF
--- a/src/cm.rs
+++ b/src/cm.rs
@@ -204,16 +204,25 @@ impl<'a, 'o> CommonMarkFormatter<'a, 'o> {
         self.need_cr = max(self.need_cr, 2);
     }
 
-    fn format_children(&mut self, node: &'a AstNode<'a>) {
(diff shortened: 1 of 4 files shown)


kivikakk:
So good! :heart: Thank you!

kivikakk:
btw, I figure this won't be super relevant since you're working off a fork, but a crate release would ever be helpful to y'all, please feel free to shout.

brson:
Thanks. It's probably worth putting this out as a point release so that crates.io and docs.rs can get it.

PR #66, opened 2018-05-06
Expose Arena

### LINKED ISSUE #62: Expose Arena
To use `parse_document`, it seems one needs to add `typed_arena` to their dependency list. Would a change to make https://github.com/kivikakk/comrak/blob/master/src/lib.rs#L101 `pub use` be ok?

Keats:
Closes #62

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,9 +68,7 @@ formatter:
 
 ``` rust
 extern crate comrak;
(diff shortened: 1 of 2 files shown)


kivikakk:
Looks great, thank you!

PR #74, opened 2018-06-05
Remove a redundant check in parse_inlines

brson:
This check is redundant with code inside `Subject::parse_inline`. It is removed in downstream snoomark because we have an additional case inside `Subject::parse_inline` that is also checking for eof, that this check interferes with.

It doesn't look to me like this check is important for either correctness or performance.


kivikakk:
Thanks!

PR #75, opened 2018-06-21
Speed ups

SSJohns:
Hey these changes are a part of a heap profiling pass that we've been working on. It basically identifies the most common things that are slowing down the parser or things that are being allocated but never used. 

Quick run-down:

876e8b9: These vecs are allocated with a certain size, but it is often the case that they are never used. We let the allocator give them size in the case that they are used.

67faebf: Often seek is 0 so we don't need to do the memory copying and overwriting unless it's different.

f9b0f9d: The reference inlines noramlly don't have a delimiter, so we start the delimiter with zero capacity.

2e00ec7: Big win, often scanners.rs creates a large reference structure when we could just check the very first character. Since these are called so often this eliminates a large amount of unneeded allocations.

Benches are done on average of 3.
Bench speedup:  `(9,511,724 + 9,335,745 + 9,226,028)/3  = 9357832`
Bench master:      `(12,177,662 + 12,865,365 + 12,590,586)/3 = 12544537`

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -1111,7 +1111,9 @@ impl<'a, 'o> Parser<'a, 'o> {
                         seeked += pos;
                     }
                 }
-                *content = content[seeked..].to_vec();
+                if seeked != 0 {
+                    *content = content[seeked..].to_vec();
+                }
                 if strings::is_blank(content) {
                     node.detach();
                 }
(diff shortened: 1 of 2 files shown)


kivikakk:
This is amazing, thank you! I'm a big fan of the scanner change; the ugliness introduced by essentially having the start of the parsing rule duplicated is unfortunate, but necessary to get these perf gains.

PR #76, opened 2018-06-27
Remove some superlinear complexity

brson:
Both `unescape` and `ltrim` were doing exponential work based on the number of matches in the input, by calling `remove` repeatedly, thus re-copying the same tail of the vector.

Surprisingly, I don't see a better safe tool for this than `rotate_left` offhand, which is a recently-stabilized API (and still isn't really the _right_ tool), so I wrote some unsafe code, and only tested it lightly.

cc @SSJohns

### DIFF
--- a/src/strings.rs
+++ b/src/strings.rs
@@ -12,7 +12,7 @@ pub fn unescape(v: &mut Vec<u8>) {
         if v[r] == b'\\' && r + 1 < v.len() && ispunct(v[r + 1]) {
             if let Some(prev) = prev {
                 let window = &mut v[(prev + 1 - found)..r];
-                window.rotate_left(found);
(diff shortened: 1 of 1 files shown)


brson:
Agh, `copy_from` is a 1.26 API too. I'll update it with an older function.

kivikakk:
Are you depending on an earlier version? I figure targeting current stable (now 1.27) is reasonable given Rust's stable backwards compatibility.

brson:
@kivikakk No I'm not, but comrak works with toolchains at least as far back as 1.23, and this doesn't seem worth breaking compatibility over.

brson:
Now it's using `ptr::copy`.

PR #77, opened 2018-06-29
Make the unicode correctness more obvious

### LINKED ISSUE #68: Parser::feed looks like it incorrectly accepts invalid unicode
Comrak is structured to assume its buffers are valid unicode - it is calling `str::from_utf8_unchecked` in multiple places, but `Parser::feed` appears to blindly accept raw byte buffers. It is though replacing `\0`, which is not a valid utf-8 byte, with the unicode replacement character, but I don't see where comrak is otherwise validating that `feed` receives utf-8.

It looks to me like if someone fed this function invalid unicode (other than `\0`) that comrak would create invalid unicode strings.

The `feed` function would probably be safer if it accpeted `&str`, punting the unicode validation to the caller. Or alternately it could validate unicode is it goes for a cost.

kivikakk:
Fixes #68.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -40,7 +40,7 @@ pub fn parse_document<'a>(
         last_line_blank: false,
     })));
     let mut parser = Parser::new(arena, root, options);
-    parser.feed(buffer.as_bytes(), true);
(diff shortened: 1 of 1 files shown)


brson:
<3

PR #80, opened 2018-07-12
Fix some link parsing cases that differ from cmark

brson:
cc @SSJohns

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -848,3 +848,25 @@ fn safe() {
         |opts| opts.safe = true,
     )
 }
+
+#[test]
+fn link_backslash_requires_punct() {
+    // Test should probably be in the spec.
+    html("[a](\\ b)", "<p>[a](\\ b)</p>\n");
+}
+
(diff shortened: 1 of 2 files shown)


src/tests.rs
```diff
@@ -848,3 +848,25 @@ fn safe() {
...
+    // Test should probably be in the spec.
+    html("[a](<<b)", "<p><a href=\"%3C%3Cb\">a</a></p>\n");
+}
+
+#[test]
+fn angle_bracketed_link_fallback_2() {
+    // When the `<...>` URL parser hits a newline it falls back to the
+    // non-bracketed parser.
+    // Test should probably be in the spec.
+    html("[a](<b\n)", "<p><a href=\"%3Cb\">a</a></p>\n");
```
kivikakk: Agree re: adding these to the spec. I'll see about getting that done, or failing that, I'll add them to the regression suite in `cmark-gfm` and double-check we're running those in CI here too.

PR #82, opened 2018-07-12
Fix unescaping of `&CounterClockwiseContourIntegral;`

brson:
Add entity_tests.py to CI.

### DIFF
--- a/src/entity.rs
+++ b/src/entity.rs
@@ -5,7 +5,7 @@ use std::cmp::min;
 use std::str;
 
 pub const ENTITY_MIN_LENGTH: usize = 2;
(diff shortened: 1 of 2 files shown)


src/entity.rs
```diff
@@ -5,7 +5,7 @@ use std::cmp::min;
 use std::str;
 
 pub const ENTITY_MIN_LENGTH: usize = 2;
-pub const ENTITY_MAX_LENGTH: usize = 31;
+pub const ENTITY_MAX_LENGTH: usize = 32;
```
kivikakk: Nice catch!

PR #83, opened 2018-07-17
Cleanup `feed` and a NUL replacement fix

brson:
My goal here was to simplify `feed` but I ended up doing a few other things:

- Fix the accounting around `\0` and newlines in the first commit - these were missed in the test suite, and seem to pass in the reference implementation.
- Remove three of the line/column accounting variables from `Ast`. Most were only read to deal with accounting from feed being called multiple times, and even then the further results not actually used meaningfully. The final one, `start_line`, is only read in one location, and removing it doesn't affect the test suite to my recollection, but messing with it isn't particularly relevant here.

### DIFF
--- a/src/nodes.rs
+++ b/src/nodes.rs
@@ -315,15 +315,6 @@ pub struct Ast {
     /// The line in the input document the node starts at.
     pub start_line: u32,
 
-    /// The column in the input document the node starts at.
-    pub start_column: usize,
-
-    /// The line in the input document the node ends at.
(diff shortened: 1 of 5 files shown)


brson:
I recognize I should upstream these test cases, though the NUL cases I think don't actually work with spec_tests.py and would need a custom test harness.

kivikakk:
Nice cleanup!

As for a custom test harness, this seems fine/likely to be accepted upstream!

PR #100, opened 2019-02-12
Callback for Broken Reference Links

sunjay:
**Wow!** This implementation is *so* much simpler than #98. I totally get why `pulldown-cmark` went with this.

---

This PR adds a new method `parse_document_with_broken_link_callback` which allows you to pass in a function/closure which will be used to resolve broken reference links. This is similar in spirit (but quite different in reality) than [the method that `pulldown-cmark` has for this](https://docs.rs/pulldown-cmark/0.2.0/pulldown_cmark/struct.Parser.html#method.new_with_broken_link_callback).

I think the approach I've taken here is even more flexible than `pulldown-cmark`. Our closure is `FnMut` instead of `Fn` which means that the user can do whatever they want in the closure to resolve the broken reference. We also use byte slices instead of string slices since that is conventionally what is used throughout the rest of this library.

The change has been tested locally and I added a doctest that both documents how to use this and to show that it really works.

I copied the additional comments I added in my previous PR to this PR so the codebase still improves even though we went with a different approach. :)

If these changes work and they get merged, feel free to close my other PR. :smile: :tada: 

## Documentation Screenshot

image

## Compatibility

Unlike my previous PR (#98), this PR is actually **completely backwards compatible**. By adding a new function and implementing the old one in terms of that, none of my changes can break anyone's code. Feel free to just bump the patch version and release this when it is ready. :tada:

### DIFF
--- a/src/parser/inlines.rs
+++ b/src/parser/inlines.rs
@@ -893,6 +893,8 @@ impl<'a, 'r, 'o, 'd, 'i> Subject<'a, 'r, 'o, 'd, 'i> {
         let is_image = self.brackets[brackets_len - 1].image;
         let after_link_text_pos = self.pos;
 
+        // Try to find a link destination within parenthesis
+
         let mut sps = 0;
         let mut url: &[u8] = &[];
         let mut n: usize = 0;
@@ -924,6 +926,8 @@ impl<'a, 'r, 'o, 'd, 'i> Subject<'a, 'r, 'o, 'd, 'i> {
                 self.pos = after_link_text_pos;
             }
         }
+
+        // Try to see if this is a reference link
 
         let (mut lab, mut found_label) = match self.link_label() {
             Some(lab) => (lab.to_vec(), true),
--- a/src/parser/inlines.rs
+++ b/src/parser/inlines.rs
@@ -15,7 +15,7 @@ use unicode_categories::UnicodeCategories;
 const MAXBACKTICKS: usize = 80;
 const MAX_LINK_LABEL_LENGTH: usize = 1000;
 
-pub struct Subject<'a: 'd, 'r, 'o, 'd, 'i> {
+pub struct Subject<'a: 'd, 'r, 'o, 'd, 'i, 'c, 'subj> {
     pub arena: &'a Arena<AstNode<'a>>,
     options: &'o ComrakOptions,
     pub input: &'i [u8],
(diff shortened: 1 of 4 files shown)


src/parser/inlines.rs
```diff
@@ -29,6 +29,10 @@ pub struct Subject<'a: 'd, 'r, 'o, 'd, 'i> {
     special_chars: [bool; 256],
     skip_chars: [bool; 256],
     smart_chars: [bool; 256],
+    // Need to borrow the callback from the parser only for the lifetime of the Subject, 'subj, and
+    // then give it back when the Subject goes out of scope. Needs to be a mutable reference so we
+    // can call the FnMut and let it mutate its captured variables.
+    callback: Option<&'subj mut &'c mut dyn FnMut(&[u8]) -> Option<(Vec<u8>, Vec<u8>)>>,
```
kivikakk: jaw drop

sunjay: Heh. I totally agree that this super ugly! I spent a long time staring at it trying to figure out if it could be made nicer. Unfortunately, it can't because of two things:

1. we need a reference to the closure because you can only have one instance of a mutable reference at any given time. You can futher borrow that reference (as we have done here), and the only other alternative is moving the reference around everywhere which is far more ugly!
2. you can only use a *mutable* reference through another *mutable* reference. You can't use it through an *immutable* reference. So `&'a &'b mut` would not work.

It's ugly, but it gives us memory safety so I guess it's okay! :sob: :laughing:

sunjay:
Thank you! :heart: I'm really excited to see these changes get released! :tada: 

I thought about adding this to the options struct, but I feel like that would only pass on the ugliness to the user (if they store the options) instead of making it any easier to us. This way, the user doesn't need to worry about any of the lifetime madness, we take care of it all internally and they just have to pass a simple callback to our function. :)

kivikakk:
Your reasoning is super solid -- thank you!

I'll make the point release now 👍

kivikakk:
Done! https://crates.io/crates/comrak/0.4.2

Thank you so much for helping out. 🙏

sunjay:
Thank you for being so responsive and helping me when I needed it! :smile: :tada:

brson:
Amazing work you've been doing @sunjay ! Kudos.

sunjay:
Thank you @brson! Happy to help out! :smile:

PR #102, opened 2019-02-12
make_block -> Ast::new, expose

kivikakk:
Ref @sunjay's comment in #101; make this function accessible and straightforward to use. (Most users won't be assigning line numbers when constructing ASTs manually.)

### DIFF
--- a/src/nodes.rs
+++ b/src/nodes.rs
@@ -362,14 +362,16 @@ pub struct Ast {
     pub last_line_blank: bool,
 }
 
-/// Create a new AST node with the given value.
-pub fn make_ast(value: NodeValue) -> Ast {
-    Ast {
-        value: value,
-        content: vec![],
-        start_line: 0,
-        open: true,
-        last_line_blank: false,
+impl Ast {
+    /// Create a new AST node with the given value.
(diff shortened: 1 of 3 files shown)


sunjay:
This is great! Thanks for adding it! What do you think about making this a method on `Ast`?

```rust
impl Ast {
    pub fn new(value: NodeValue) -> Self {
        //...
    }
}
```

I think this is probably closer to what Rust programmers typically expect. :)

kivikakk:
Of course! I'm too used to the codebase to notice this kind of mistake; thanks for the pickup!

kivikakk:
My pleasure! Thanks for bringing these kinds of changes to my attention. I'm not currently an active user of the library so having people who _are_ using it participate in its development is the best! 👍

sunjay:
Hi @kivikakk! Could you release a new patch version when you have the opportunity? I'd love to get these latest changes released so I can start using them. :smile:

kivikakk:
@sunjay Of course, done! https://crates.io/crates/comrak/0.4.3

sunjay:
Thank you very much. :heart_eyes: :smile:

PR #105, opened 2019-02-22
Added a constructor for AstNode

sunjay:
Hello again!
I added a constructor for AstNode so we don't have to go through 3 constructors every time just to create one.

The issue I'm addressing is that `AstNode` is `Node<'a, RefCell<Ast>>` which is three nested objects. That means to make one from a `NodeValue`, you have to use three constructors:

```rust
Node::new(RefCell::new(Ast::new(value)))
```

Rather than just make a utility function in my code, I thought it would be better to just add a constructor so you can do `AstNode::from(value)` instead.

If/when this gets merged, could you release a new patch version of `comrak`? Thanks as always! :smile: :tada: 

## Documentation Screenshot

image

### DIFF
--- a/src/nodes.rs
+++ b/src/nodes.rs
@@ -380,7 +380,21 @@ impl Ast {
 /// It is bound by the lifetime `'a`, which corresponds to the `Arena` nodes are allocated in.
 /// `AstNode`s are almost handled as a reference itself bound by `'a`.  Child `Ast`s are wrapped in
 /// `RefCell` for interior mutability.
+///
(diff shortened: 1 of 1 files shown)


kivikakk:
Done -- https://github.com/kivikakk/comrak/releases/tag/0.4.4.

sunjay:
Thank you so much! 😊

kivikakk:
Just FYI (and I'm not sure this'll be relevant given how much time has passed), I'm planning to remove this `From` impl in 0.18 — with proper line/col tracking to come, I'm discouraging creating `Ast`s with default (zero) values for these fields.

PR #110, opened 2019-03-15
Make FootnoteDefinition names match FootnoteReference one

### LINKED ISSUE #109: Content of AST for FootnoteReference and FootnoteDefinition is inconsistent
If you have a footnote like this: 

```
Some footnote[^foo]

[^foo]: content of the footnote
```
In `FootnoteReference` the content of the `Vec[u8]`payload will be a number (e.g. `1`) while in `FootnoteDefinition` the payload content will be the literal (i.e. `foo`). This makes associating the two a bit tricky.

lise-henry:
This sould fix #109. Not sure if this is the right way to do it but it seems to work.


kivikakk:
LGTM 👍  Let me know if you want a new release done.

PR #115, opened 2019-04-23
Add TaskItem variant to NodeValue

### LINKED ISSUE #108: Checkboxes only work in HTML
Currently, checkboxes lists are (as far as I understand) are lists where the `[ ]` and `[X]` are replaced by some HTML code. While this works for HTML outputs this prevents this features being used in other output (e.g. LaTeX in my use case), while most other features work really pretty nicely. Do you think it would be possible to modify the AST instead of the content of the list item? Since list items have a `NodeList` elements I suppose it would be possible to add some fields to distinguish between 'normal' list items and the two variants of checkboxes without breaking compatibility.

Anyway, thanks for the work :)

lise-henry:
This enables to only generate HTML at the time of rendering. Since this adds a variant this is a breaking change.

Fixes #108


kivikakk:
This is awesome! Thank you so much!

kivikakk:
Released in https://github.com/kivikakk/comrak/releases/tag/0.6.0!

Issue #116, opened 2019-04-30
Feature Requests: Command line Flag to turn on all common extensions?


jrmiller82:
Could we get a `-a` or `--all` to keep us from having to type `-e foo` over and over for each feature? Like a flag that turns on all extensions except for my header tags and soft links become hard links?

kivikakk:
I think this is not a good idea. As more extensions are added, anyone using such a flag would have a surprising result when upgrading as the meaning of their documents changes.

I'm not sure what you mean by "my header tags and soft links become hard links", but if you already want to exclude extensions, that suggests maybe enabling all extensions at once isn't a good idea, and that we would also need to implement options to then disable already-enabled extensions. The complexity burden for a convenience option is undesirable.

If you find yourself typing out the list of extensions a lot, you might need to instead consider what your process is that leads to this, and whether or not you could reduce that burden another way.

jrmiller82:
I didn't realize that the soft line breaks become hard line breaks weren't a `-e foo` extension but a separate flag. What about a CLI flag just for the github-style extensions? Like `--git` or something? I completely see what you're saying, but the git extensions haven't changed in a long time and likely won't, right?

kivikakk:
That's a good point. I think the best design would be a `--gfm` flag that enables the GFM extensions, which is to say `strikethrough`, `tagfilter`, `table`, `autolink`, and `tasklist`. It should also enable `--github-pre-lang`.

I'm happy to do this, but would also like to offer writing the PR to you in case it's something you'd like to do.

jrmiller82:
I can take a first stab at it. My rust-fu is still novice to intermediate.

jrmiller82:
Should have you a PR tonight.

jrmiller82:
PR #118 up. I hope it helps. I apologize for my first rambling feature request; I used incorrect vocabulary and I apologize for any wasted time. I hope the PR helps with UI/UX on the CLI. :)

kivikakk:
Thank you so much, that's been merged! And nothing to apologise for, no time's wasted when communicating. :)

Let me know if you'd like a new release of comrak to be made with these new changes.

jrmiller82:
You're welcome! I have no need for an immediate release if you'd like to wait until your regularly planned release schedule. :)

kivikakk:
Haha, I ask in part because there's no regular cadence at present. I'm a real fan of this change so I'll make a release now!

jrmiller82:
[...]

PR #118, opened 2019-05-01
create new flag --gfm

jrmiller82:
Fixes Issue #116. However, I didn't see if you had implemented any tests for the flags so I didn't write any tests. I also did not run rustfmt as I saw it would change another unrelated code block in src/main.rs. 

Hope this helps. :)


jrmiller82:
Added a second commit to add me to the contributors.txt file.

kivikakk:
Thank you! This looks great. There aren't any flag tests yet, but that's something we should consider changing as we add more flags.

Worth noting, if you type "fixes #116" in the PR text body (without any word in between), the issue will automatically get closed when the PR is merged.

PR #135, opened 2020-01-20
Remove list item indentation

casey:
I'm using comrak to parse and dump a CM readme so I can apply automatic modifications to it. (In particular, adding a table of contents and keeping a table derived from an external data-source up to date.)

I noticed that list items get two extra spaces of indentation when formatted as CM. I think perhaps this is a bug, but I wasn't sure.

Since it's a relatively simple change to remove it, I decided to open a PR instead of opening an issue.

Please feel free to close if it's intentional!

### DIFF
--- a/src/cm.rs
+++ b/src/cm.rs
@@ -323,7 +323,7 @@ impl<'a, 'o> CommonMarkFormatter<'a, 'o> {
                 let mut listmarker = vec![];
 
                 let marker_width = if parent.list_type == ListType::Bullet {
-                    4
+                    2
                 } else {
                     let mut list_number = parent.start;
                     let list_delim = parent.delimiter;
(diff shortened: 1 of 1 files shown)


kivikakk:
Thanks for this! I'll check it over in the morning. :) I think CI for PRs is broken too, heh.

kivikakk:
Good news — CI is working again! The bad news is it's broken, but I'm looking now into a fix.

kivikakk:
So, to answer the original question: it's not a bug, per se, because it's valid CommonMark that represents the same syntax tree as the original input, but it is more spacing than we need, and I prefer keeping the output minimal, so your PR is appreciated 👍 

Once CI passes on my follow up changes, will merge!

casey:
Sweet!

kivikakk:
Lemme know if you want a new release for the change.

casey:
Thank you! For now I don't, but if that changes I'll let you know.

PR #150, opened 2020-08-11
Add option to escape HTML instead of clobber (fix #149)

### LINKED ISSUE #149: Escape HTML instead of clobbering?
Does/will comrak support escaping HTML, so it displays in a browser as typed, rather than clobbering?

yujiri8:
This is my first time contributing to a Rust library :)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -57,6 +57,11 @@ fn main() -> Result<(), Box<dyn Error>> {
                 .help("Allow raw HTML and dangerous URLs"),
         )
         .arg(
+            clap::Arg::with_name("escape")
+                .long("escape")
+                .help("Escape raw HTML instead of clobbering it"),
+        )
+        .arg(
             clap::Arg::with_name("extension")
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -134,6 +134,7 @@ fn main() -> Result<(), Box<dyn Error>> {
                 .parse()
                 .unwrap_or(0),
             unsafe_: matches.is_present("unsafe"),
+            escape: matches.is_present("escape"),
```
kivikakk, line 137: We'll need to add a corresponding `clap::Arg` above to get this option from the command line.

yujiri8: I've updated the commit to include command-line support.

kivikakk:
Looking good! Will merge and prepare a new release.

kivikakk:
The release is done! Your changes are now part of [comrak 0.8.1](https://crates.io/crates/comrak). Thank you so much! :heart:

PR #152, opened 2020-08-12
Fix Clippy lint warning

alerque:
Making Clippy happy is a good thing, cluttering build output with lint warnings is a bad thing.


kivikakk:
Thanks!

PR #154, opened 2020-08-12
fix up how we invoke benchmarks

### LINKED ISSUE #151: Latest releases lost support for Rust stable
From the readme:

> Comrak supports Rust stable.

This is not true of the latest rush of releases (0.7.1, 0.8.0, or 0.8.1). Here is the Clippy error:

```console
error[E0554]: `#![feature]` may not be used on the stable release channel
  --> src/lib.rs:74:37
   |
74 | #![cfg_attr(feature = "benchmarks", feature(test))]
   |                                     ^^^^^^^^^^^^^
```

Rustup informs me I'm running:

    stable-x86_64-unknown-linux-gnu unchanged - rustc 1.45.2 (d3fb005a3 2020-07-31)

Please either fix this so stable Rust can be used again or change the documentation to reflect the actual state of affairs (and maybe add a note about the change in the release notes for the sake of package builders).

kivikakk:
Don't make benchmarks a "feature" of a regular test run; put the benchmark in the `benches` directory so it's invoked with `cargo bench`.

(Awkwardly, note that `cargo bench` won't actually work on stable either—use `cargo +nightly bench` or similar.)

Fixes #151.
/cc @alerque

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,4 +36,3 @@ timebomb = "0.1.2"
 
 [features]
 default = ["clap"]
-benchmarks = []
--- a/benches/progit.rs
+++ b/benches/progit.rs
@@ -1,0 +1,24 @@
+#![feature(test)]
+
+extern crate test;
+extern crate comrak;
+
+use comrak::{parse_document, format_html, Arena, ComrakOptions};
+use test::Bencher;
+
+#[bench]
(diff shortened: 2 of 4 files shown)


alerque:
This seems like a reasonable improvement at least until the `#[bench]` attribute hits stable.

kivikakk:
Great. Will merge.

PR #157, opened 2020-08-13
add a config file at ~/.config/comrak/config

### LINKED ISSUE #155: .comrakrc for cli default options?
I think the most common use case for the CLI would be to do something like this:

```bash
comrak --unsafe --gfm --github-pre-lang --header-ids '' README.md
```

However, that's quite a bit to type.

Do you think it would be good to add the functionality of reading default options from `$HOME/.config/comrak/config` or `$HOME/.comrakrc`?

If not, I suppose that an alias such this would be a good workaround:

```bash
alias comrak="comrak --unsafe --gfm --github-pre-lang --header-ids ''"
```

kivikakk:
Fixes #155.

/cc @solderjs — what do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,11 +45,14 @@ FLAGS:
         --github-pre-lang    Use GitHub-style <pre lang> for code blocks
         --hardbreaks         Treat newlines as hard line breaks
     -h, --help               Prints help information
+        --no-config-file     Do not read config file
         --smart              Use smart punctuation
         --unsafe             Allow raw HTML and dangerous URLs
     -V, --version            Prints version information
 
 OPTIONS:
+    -c, --config-file <PATH>            Path to config file containing command-line arguments [default:
+                                        /Users/kivikakk/.config/comrak/config]
         --default-info-string <INFO>    Default value for fenced code block's info strings if none is given
     -e, --extension <EXTENSION>...      Specify an extension name to use [possible values: strikethrough, tagfilter,
                                         table, autolink, tasklist, superscript, footnotes, description-lists]
@@ -59,6 +62,9 @@ OPTIONS:
 
 ARGS:
     <FILE>...    The CommonMark file to parse; or standard input if none passed
+
+By default, Comrak will attempt to read command-line options from a config file specified by --config-file.  This
+behaviour can be disabled with --no-config-file.  It is not an error if the file does not exist.
 ```
 
 And there's a Rust interface. You can use `comrak::markdown_to_html` directly:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -45,11 +45,14 @@ FLAGS:
...
         --hardbreaks         Treat newlines as hard line breaks
     -h, --help               Prints help information
+        --no-config-file     Do not read config file
         --smart              Use smart punctuation
         --unsafe             Allow raw HTML and dangerous URLs
     -V, --version            Prints version information
 
 OPTIONS:
+    -c, --config-file <PATH>            Path to config file containing command-line arguments [default:
+                                        /Users/kivikakk/.config/comrak/config]
```
solderjs, line 55: Perhaps the documentation should show `~/.config/comrak/config`? Or is this dynamic?

README.md
```diff
@@ -45,11 +45,14 @@ FLAGS:
         --github-pre-lang    Use GitHub-style <pre lang> for code blocks
         --hardbreaks         Treat newlines as hard line breaks
     -h, --help               Prints help information
+        --no-config-file     Do not read config file
```
solderjs, line 48: What happens in the case of `--no-config-file --config-file ./config`? I assume `--no-config-file` takes precedence? Would it be just as well to have `--config-file false`?

src/main.rs
```diff
@@ -173,3 +220,15 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     process::exit(0);
 }
+
+fn get_default_config_path() -> String {
```
solderjs, line 224: Looking at https://github.com/whitequark/rust-xdg/blob/master/src/lib.rs I thought that this might not handle windows, but following through to https://crates.io/crates/dirs, it looks like it is cross-platform after all.

README.md
kivikakk: The output is dynamic, yeah — it'll show the actual default based on their XDG config path. The README excerpt is generated from running the actual binary on my machine, so it shows mine as a sample. :)

src/main.rs
kivikakk: Nice! Rust libraries have been pretty good at Windows compat in my experience.

README.md
kivikakk: Yep, `--no-config-file` takes precedence. Hesitant to use `false` since it's not a boolean per se, but I do like not having the extra option, so will make `none` recognised!

kivikakk:
Awesome, thanks for your careful review!

I'll do a release shortly.

src/main.rs
kivikakk: Oh, actually. Nope. >_< Missed the first line:

```rust
#![cfg(any(unix, target_os = "redox"))]
```

This doesn't work on Windows at all.

kivikakk: We could do something like this: https://github.com/DanielKeep/cargo-script/blob/614e60e5932e218ebff1e471303eb0d59870d03b/src/platform.rs#L309-L363

But for now I'm defaulting to `none` on Windows.

PR #165, opened 2020-08-30
add webinstall.dev install instructions

solderjs:
I was going to propose this earlier, but I needed to get a permissions issue with Catalina fixed.

Now it installs without asking for Admin permissions, even though the binary isn't signed.


kivikakk:
Nice, thank you!

(I've removed the "Build from Source" heading as it's not an instruction for building from source, but rather using as a library in one's own software, as the primary use case.)

PR #169, opened 2020-10-29
Allow "image/" prefix on data URIs

dancodes:
Update the `dangerous_url` grammar rule to allow URIs that begin with `data:image/`, followed by a valid media type. This is an optional modifier and does not affect URIs that do not contain this prefix.

I'd like to add this because this is the default behavior of [Toast UI's Editor](https://ui.toast.com/tui-editor/) when adding images and outputting Markdown code.

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs), Data URIs consist of the following format:

```
data:[<mediatype>][;base64],<data>
```

This documentation then states:

> The mediatype is a MIME type string, such as 'image/jpeg' for a JPEG image file.

### DIFF
--- a/src/lexer.pest
+++ b/src/lexer.pest
@@ -55,4 +55,4 @@ table_start = { "|"? ~ table_marker ~ ("|" ~ table_marker)* ~ "|"? ~ table_space
 table_cell_end = { "|" ~ table_spacechar* ~ table_newline? }
 table_row_end = { table_spacechar* ~ table_newline }
 
-dangerous_url = { "data:" ~ !("png" | "gif" | "jpeg" | "webp") | "javascript:" | "vbscript:" | "file:" }
+dangerous_url = { "data:" ~ "image/"? ~ !("png" | "gif" | "jpeg" | "webp") | "javascript:" | "vbscript:" | "file:" }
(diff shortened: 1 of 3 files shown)


dancodes:
Nightly build seems to have failed due to missing `rustfmt` binary:

```error: component 'rustfmt' for target 'x86_64-unknown-linux-gnu' is unavailable for download for channel nightly```

File: src/lexer.pest [hunk 1]
```diff
@@ -55,4 +55,4 @@ table_start = { "|"? ~ table_marker ~ ("|" ~ table_marker)* ~ "|"? ~ table_space
 table_cell_end = { "|" ~ table_spacechar* ~ table_newline? }
 table_row_end = { table_spacechar* ~ table_newline }
 
-dangerous_url = { "data:" ~ !("png" | "gif" | "jpeg" | "webp") | "javascript:" | "vbscript:" | "file:" }
+dangerous_url = { "data:" ~ "image/"? ~ !("png" | "gif" | "jpeg" | "webp") | "javascript:" | "vbscript:" | "file:" }
```
kivikakk, line 58:
```suggestion
dangerous_url = { "data:" ~ !("image/" ~ ("png" | "gif" | "jpeg" | "webp")) | "javascript:" | "vbscript:" | "file:" }
```

kivikakk (approved):
Thanks! This was a mistranslation from the original: https://github.com/commonmark/cmark/blob/8410d2c2ae1c5ec2cbde51e87d81ba8ef6898844/src/scanners.re#L300

It should read: a URL is dangerous if it is a `data:` URI that is not followed by `image/png`, `image/gif`, etc. There's no need to allow `data:png` or similar -- that was just mistaken.

kivikakk:
> Nightly build seems to have failed due to missing `rustfmt` binary:

I'll fix this up when I'm next merging changes -- thank you!

PR #170, opened 2020-12-06
Add support for ignoring front matter.

emosenkis:
This is a working proof of concept. If you're amenable to merging it, I can address any comments about the design/implementation and add unit tests (right now, there's only two doctests).

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -480,20 +480,14 @@ impl<'a, 'o, 'c> Parser<'a, 'o, 'c> {
 
         if let Some(ref delimiter) = self.options.extension.front_matter_delimiter {
             let front_matter_pattern = RegexBuilder::new(&format!(
-                "\\A(?:\u{feff})?{delim}$.*^{delim}$",
+                "\\A(?:\u{feff})?{delim}\\r?\\n.*^{delim}\\r?\\n(?:\\r?\\n)?",
                 delim = regex::escape(delimiter)
             ))
             .multi_line(true)
             .dot_matches_new_line(true)
             .build()
             .unwrap();
-            if let Some(mut front_matter_size) = front_matter_pattern.shortest_match(s) {
-                if let Some(13 /* \r */) = s.get(front_matter_size) {
-                    front_matter_size += 1;
-                }
-                if let Some(10 /* \n */) = s.get(front_matter_size) {
-                    front_matter_size += 1;
-                }
+            if let Some(front_matter_size) = front_matter_pattern.shortest_match(s) {
                 i += front_matter_size;
                 let node = self.add_child(self.root, NodeValue::FrontMatter(s[..i].to_vec()));
                 self.finalize(node).unwrap();
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -10,7 +10,7 @@ use nodes::{
     Ast, AstNode, ListDelimType, ListType, NodeCodeBlock, NodeDescriptionItem, NodeHeading,
     NodeHtmlBlock, NodeList, NodeValue,
 };
-use regex::bytes::Regex;
+use regex::bytes::{Regex, RegexBuilder};
 use scanners;
 use std::cell::RefCell;
 use std::cmp::min;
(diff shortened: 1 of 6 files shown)


src/parser/mod.rs
```diff
@@ -435,8 +475,31 @@ impl<'a, 'o, 'c> Parser<'a, 'o, 'c> {
...
+                    front_matter_size += 1;
+                }
+                if let Some(10 /* \n */) = s.get(front_matter_size) {
+                    front_matter_size += 1;
+                }
+                i += front_matter_size;
+                let node = self.add_child(self.root, NodeValue::FrontMatter(s[..i].to_vec()));
+                self.finalize(node).unwrap();
+            }
+        }
```
kivikakk: I don't love using regular expressions to do this -- I'd much rather do some string slice ops and avoid the somewhat messy regex format/compile/match on a possibly huge string. This is just a style thing, though, unless someone can find a way to break it.

I will push one patch here, which allows use of `\r\n` as a linebreak (unfortunately [`$` only matches `\n`](https://github.com/rust-lang/regex/issues/244)), and collapses the manual logic stepping afterward. We still use `^` to match the start of the second delimiter, since it'll still work for our purposes here.

Needing to compile a regex per parse is suboptimal in a \~feelsy~ way, but in practice it makes negligible difference to performance to documents that don't include frontmatter, and so from that perspective I'm currently unconcerned.

#### comrak `main`

```shell
raven ~/k/comrak (main|✔) $ cargo build --release
    Finished release [optimized] target(s) in 0.17s
raven ~/k/comrak (main|✔) $ make -C vendor/cmark-gfm/ bench PROG=../../target/release/comrak
make: Entering directory '/home/kivikakk/k/comrak/vendor/cmark-gfm'
{ for x in `seq 1 20` ; do \
        /usr/bin/env time -p ../../target/release/comrak </dev/null >/dev/null ; \
        /usr/bin/env time -p ../../target/release/comrak bench/benchinput.md >/dev/null ; \
        done \
} 2>&1  | grep 'real' | awk '{print $2}' | python3 'bench/stats.py'
mean = 1.1230, median = 1.1150, stdev = 0.0360
make: Leaving directory '/home/kivikakk/k/comrak/vendor/cmark-gfm'
```

#### frontmatter (with option enabled)

```shell
raven ~/k/comrak (emosenkis-main|●1) $ cargo build --release
   Compiling comrak v0.9.0 (/home/kivikakk/k/comrak)
    Finished release [optimized] target(s) in 27.72s
raven ~/k/comrak (emosenkis-main|●1) $ make -C vendor/cmark-gfm/ bench PROG="../../target/release/comrak --front-matter-delimiter ~~~"
make: Entering directory '/home/kivikakk/k/comrak/vendor/cmark-gfm'
{ for x in `seq 1 20` ; do \
        /usr/bin/env time -p ../../target/release/comrak --front-matter-delimiter ~~~ </dev/null >/dev/null ; \
        /usr/bin/env time -p ../../target/release/comrak --front-matter-delimiter ~~~ bench/benchinput.md >/dev/null ; \
        done \
} 2>&1  | grep 'real' | awk '{print $2}' | python3 'bench/stats.py'
mean = 1.1275, median = 1.1250, stdev = 0.0112
make: Leaving directory '/home/kivikakk/k/comrak/vendor/cmark-gfm'
```

kivikakk (approved):
Hey, thank you so much for this! Sorry it took me so long to get back to you, life's been super busy lately. I have some comments (below) and two commits I've pushed; we needed to tell Clap to support values with leading hyphens so we could specify e.g. `--front-matter-delimiter ---`. This looks good to go!

kivikakk:
If you'd like to submit a follow up PR with more test coverage, I'd gladly accept it! But given the limited scope of the change, what's here is sufficient for release. Let me know if you'd like me to make a release.

src/parser/mod.rs
emosenkis: Yeah, I tried to do manual string slicing at first, but by the time I finished handling Unicode byte order markers and multiple forms of line endings, I would have ended up with a pretty sizable state machine, which is exactly what the regexp should be doing anyway. While it's unfortunate to compile the regexp every time instead of at build time, the simplicity of the regexp should ensure that it compiles very quickly. Once compiled, the regexp should be roughly as efficient as a hand-written state machine.

PR #174, opened 2021-02-15
Builders for options

kivikakk:
I keep breaking semver compatibility by adding options. Change public-facing options to builders so adding an option is backwards-compatible.

I'm not super sold on this variety of builder, namely the way you need to call `options.x` for several `x`s if you want to modify multiple kinds. Maybe I could roll all the builder functions into one impl.

### DIFF
--- a/examples/s-expr.rs
+++ b/examples/s-expr.rs
@@ -16,7 +16,7 @@ const INDENT: usize = 4;
 const CLOSE_NEWLINE: bool = false;
 
 use comrak::nodes::{AstNode, NodeValue};
-use comrak::{parse_document, Arena, ComrakExtensionOptions, ComrakOptions};
+use comrak::{parse_document, Arena, ComrakOptions};
 use std::env;
 use std::error::Error;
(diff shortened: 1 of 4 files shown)


kivikakk:
@YJDoc2 Sorry I never replied to your comment! Life got a lot.

I'd be very happy to accept such a PR. Let me know if this is something you still have any interest in taking on — and I understand it might not be, of course, given the passage of time — otherwise I'll likely get to it soon while tidying up Comrak.

YJDoc2:
Hey, no worries! :smile: 

I am still interested in helping out, but I'll need to take another look at what I was suggesting as I seem to have forgotten that myself :joy: 

Also, I am a bit busy with other stuff at the moment as well, so how about this :
I will try to get a PR opened up as I can, that adds the suggested builder stuff ; but in case you get to tidying up before it, feel free to continue with it and close this as appropriate. 

In case you need a more concrete timeline, let me know, I'll try to get it done sooner, or just pass on it :+1:

PR #189, opened 2021-06-23
refactor: extract arms of format_node into functions

smelm:
I am currently trying to understand the codebase in order to make some real contributions down the line (I'm really excited about the project). While I'm at it I might as well do some small refactorings to make the code easier to understand for myself and the next person who reads it. 

In this one I just extract the arms of a big match statement into functions which makes it easier to see in one glance.

Please let me know if those kinds of PRs are a welcome or more of a hassle on your end.

### DIFF
--- a/src/cm.rs
+++ b/src/cm.rs
@@ -1,7 +1,10 @@
 use ctype::{isalpha, isdigit, ispunct, isspace};
 use nodes;
 use nodes::TableAlignment;
-use nodes::{AstNode, ListDelimType, ListType, NodeLink, NodeValue};
+use nodes::{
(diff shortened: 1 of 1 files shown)


kivikakk:
Looks fine to me! The cargo fmt step failed so I just ran that to get CI green again.

In general, I'm happy to take refactorings on board; the main thing I'd like to avoid is significantly changing how the logic works, to preserve its close mapping of upstream [cmark-gfm](https://github.com/github/cmark), so we can closely track spec changes by mapping the upstream's implementation's changes.

kivikakk:
Thank you very much! :heart: I'll look forward to any further contributions you might have 😄 

I should note — I am not currently a user of comrak. It's been a couple years since I stopped working on something that used it. But it is used in some notable places, like docs.rs and crates.io.

PR #191, opened 2021-06-28
Fix bug where autolink fails to recognize short URLs with non-empty paths

### LINKED ISSUE #190: Single level domain names are not recognized by the autolink extension
When autolink is enabled (`options.extension.autolink = true;`), text that look like URLs should be converted into links.

This works, as expected: `http://example.com/path/to/my/page`

However, this doesn't: `http://example/path/to/my/page`

The CommonMark spec doesn't seem to require that the domain name must have at least two levels.  However, a single level domain name is a valid URL, and it's common in private networks.

btzy:
Resolves #190.


kivikakk:
Per https://github.com/kivikakk/comrak/issues/190#issuecomment-874700727, this LGTM :shipit:

PR #192, opened 2021-07-14
Add the NodeCode struct to expose number of backticks from inline code spans

vojtechkral:
I would like to use the information how many backtics are used for a code span,
ie. being able to tell whether `` `foo` `` or ` ``foo`` ` was used.
This adds the code to expose that information in the API.

I'm not too happy about it being a breaking change, but didn't see how it could be done otherwise.
Let me know if this is acceptable at all...

Edit: Test included.

Thanks for a nice md parser :slightly_smiling_face:

### DIFF
--- a/examples/headers.rs
+++ b/examples/headers.rs
@@ -3,7 +3,7 @@
 extern crate comrak;
 
 use comrak::{
-    nodes::{AstNode, NodeValue},
+    nodes::{AstNode, NodeCode, NodeValue},
     parse_document, Arena, ComrakOptions,
 };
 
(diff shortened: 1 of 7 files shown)


kivikakk:
@vojtechkral This is excellent -- thank you so much!

It is breaking, but that can't be helped with the API designed currently as it is. That's why we're in 0.x :)

Would you like a release for this, or are you happy to track `main` for now?

vojtechkral:
Thanks for the quick review :)
No need for a release as far as I'm concerned, tracking git's perfectly fine until there's enough reasons for a new release...

vojtechkral:
Actually, looks like I can't push my [project](https://github.com/vojtechkral/bard) to `crates.io` with a git dependency. Didn't know that was forbidden. So, if you feel like making a release in some near future that would help me. (But not a huge problem if inconvenient.)

kivikakk:
Ahh, how frustrating. I have made it happen → [0.11.0](https://github.com/kivikakk/comrak/releases/tag/0.11.0).

vojtechkral:
Brilliant, thanks!

PR #199, opened 2021-09-10
Set explicit features on syntect

lambda-fairy:
This lets us drop the YAML dependency.


lambda-fairy:
This only drops two dependencies, so feel free to close this if you think it's not worth the trouble :)

kivikakk:
Big fan! Thank you :heart:

PR #210, opened 2022-05-02
Improve handling of newline in inline code block

mikeando:
This case ``` `\n` ``` was getting parsed as `<code></code>` rather than `<code> </code>` due to the `\n` getting flagged as a non-space character. Change is pretty simple

There's actually an unrelated commit with this that changes error reporting to use `#[track_callers]` that makes test failures point to the right line (rather than inside the `compare_strings` helper function)

I can break this into two PRs if you want -- but they seemed trivial enough to review together.

### DIFF
--- a/src/strings.rs
+++ b/src/strings.rs
@@ -65,7 +65,7 @@ pub fn normalize_code(v: &[u8]) -> Vec<u8> {
             }
             c => r.push(c),
         }
-        if v[i] != b' ' {
+        if v[i] != b' ' && v[i] != b'\r' && v[i] != b'\n' {
             contains_nonspace = true;
         }
 
@@ -273,4 +273,19 @@ pub fn extract_attributes_from_tag(html_tag: &str) -> HashMap<String, String> {
     }
 
     attributes
+}
+
+#[cfg(test)]
+pub mod tests {
+    use super::normalize_code;
+
+    #[test]
+    pub fn normalize_code_handles_lone_newline() {
+        assert_eq!(normalize_code(&[b'\n']), vec![b' ']);
+    }
+
+    #[test]
+    pub fn normalize_code_handles_lone_space() {
+        assert_eq!(normalize_code(&[b' ']), vec![b' ']);
+    }
 }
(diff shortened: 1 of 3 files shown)


src/strings.rs
```diff
@@ -63,10 +63,10 @@ pub fn normalize_code(v: &[u8]) -> Vec<u8> {
             b'\n' => {
                 r.push(b' ');
             }
-            c => r.push(c),
-        }
-        if v[i] != b' ' {
-            contains_nonspace = true;
+            c => {
+                r.push(c);
+                contains_nonspace = true;
```
mikeando, line 68: This obviously is missing an `if c!=b' ' { contains_nonpace = true; }`.
That's why the testis failing. Sorry.

How do I run the spec tests manually?

mikeando:
Ugh, there is something trickier going on than I'd thought.
The same node ends up getting processed for inlines more than once.
It will be a day or so before I have time to dig through what is going on...
Changing this PR to WIP until then.

src/strings.rs
mikeando: Found the `cibuild` script which contained the commands to run the spec tests.

```
python3 spec_tests.py --program='../../../target/debug/comrak --syntax-highlighting none'
```

mikeando:
The trouble I was having was actually with the round-trip tests, not where I thought it was.
They're fixed too now I think. And all the spec tests pass on my machine.

The cause ofmy confusion was that I didn't realise the `html` function was doing more than just 
checking the output matched the provided HTML (and actually does an additional round-trip check).
I then added a bunch of debug prints that sent me down the wrong path altogether. Spent a long time 
trying to understand why my Code node was getting processed twice... with different content.

Anyway,  sorry about the delay on this.

sarahirbis:
ta, this looks superb! I ran `cargo fmt` to clear up the last check and pushed. It looks like this change mighta crept in between the last two CM spec versions — at some point we should actually make sure we're compliant with [0.30](https://spec.commonmark.org).

sarahirbis:
if you want a release with these changes, give us a shout, otherwise we'll just wait until there's more accumulated.

mikeando:
No need for a new release. I'm happy to point my cargo at the commit for now.

Issue #213, opened 2022-05-14
parsing of  hard-linebreak then space inside block-quote: `>\\\n A`


mikeando:
The markdown dingus (and my alternate implementation) says that

```
>\
 A
```

should parse as 

```
  <block_quote>
    <paragraph>
      <linebreak />
      <text>A</text>
    </paragraph>
  </block_quote>
```

And gives the HTML

```
<blockquote>
<p><br />
A</p>
</blockquote>
```

Note that there is no space before the `A`.

In contrast, `comrak` retains the space before the A, giving the HTML:

```
<blockquote>
<p><br />
 A</p>
</blockquote>
```

I think the correct behaviour is to remove the leading space as the spec contains:

> 6.7Hard line breaks
> ...
> Leading spaces at the beginning of the next line are ignored:

mikeando:
I'm happy to see if this is going to be easy to fix if you think it is worthwhile.

mikeando:
Seems that the issue is restricted to the `\\\n` kind of hard-linebreak. The `␣␣\n` kind seems to work fine.
Also, it is not due to the otherwise empty first line.

```
> A\
␣B
```

generates the leading space,

```
<blockquote>
<p>A<br />
␣B</p>
</blockquote>
```

while 

```
> A␣␣
␣B
```

does not generate the leading space.

```
<blockquote>
<p>A<br />
B</p>
</blockquote>
```

mikeando:
Similarly, the non-lazy form works as expected too.

```
```
>␣A\
>␣B
```

gives no extra space

```

```
<blockquote>
<p>A<br />
B</p>
</blockquote>
```

kivikakk:
Closed by #214 — thanks!

PR #214, opened 2022-05-16
Remove leading spaces after linebreak

mikeando:
There are three independent commits

The initial commit just tidies up `parse_inline` a little by:
* assigning directly to `new_inl`as the return value from the `match` rather than inside each branch.
* moving some `if c=='X' && something() {...}` conditions inside the default match arm into their own arms. 

The second commit change the way hard linebreaks were rendered in the CM output from '␣␣\n' to '\\\n' to allow
for round-tripping when the LineBreak was on an otherwise empty line. 

The final commit just left-trims the content in `parse_inline` if the last node was a `LineBreak`.
This removes the extra space character that was found in #213.

I'm not 100% sure this is the right way to fix it, but all the `cargo test` tests pass, and the spec tests all parse.

### DIFF
--- a/src/cm.rs
+++ b/src/cm.rs
@@ -523,7 +523,7 @@ impl<'a, 'o> CommonMarkFormatter<'a, 'o> {
     fn format_line_break(&mut self, entering: bool) {
         if entering {
             if !self.options.render.hardbreaks {
-                write!(self, "  ").unwrap();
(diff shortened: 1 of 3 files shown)


kivikakk:
Code changes are good, tested it out a bit locally, looks great — thanks!

Issue #215, opened 2022-05-18
Make private fields public to enable programmatically generating an AST


emschwartz:
First of all, thank you for creating and maintaining this library!

I have a use case where I would like to convert between a custom format and markdown (in both directions) and I would like to use comrak for this.

I have the conversion from markdown working but converting to markdown with comrak isn't fully possible because a number of comrak structs have crate-private fields that make it impossible for other crates to construct a comrak AST.

Would you accept a PR to make the following fields and functions public?
- [`NodeCodeBlock.fence_offset`](https://github.com/kivikakk/comrak/blob/7b1f19f2be0de1abf26dbf93ddcba9b653e11fae/src/nodes.rs#L263)
- [`NodeList.marker_offset`](https://github.com/kivikakk/comrak/blob/7b1f19f2be0de1abf26dbf93ddcba9b653e11fae/src/nodes.rs#L195)
- [`NodeList.padding`](https://github.com/kivikakk/comrak/blob/7b1f19f2be0de1abf26dbf93ddcba9b653e11fae/src/nodes.rs#L196)
- [`NodeHtmlBlock.block_type`](https://github.com/kivikakk/comrak/blob/7b1f19f2be0de1abf26dbf93ddcba9b653e11fae/src/nodes.rs#L288)
- [`can_contain_type`](https://github.com/kivikakk/comrak/blob/7b1f19f2be0de1abf26dbf93ddcba9b653e11fae/src/nodes.rs#L411)

kivikakk:
Hiya! Likewise, thanks for your interest in contributing!

> Would you accept a PR to make the following fields and functions public?

I would! Go for it.

PR #216, opened 2022-05-18
Make node value fields public

emschwartz:
Resolves https://github.com/kivikakk/comrak/issues/215

Please double-check that the field descriptions are correct. I had a bit of a hard time figuring out what each of them do.


kivikakk:
Thank you! These all look correct (enough?!) to me. They're similarly undocumented upstream: https://github.com/commonmark/cmark/blob/8a023286198a7e408398e282f293e3b0baebb644/src/node.h#L16

PR #219, opened 2022-05-29
disable control characters in link definitions

### LINKED ISSUE #218: incorrect parsing of link definition with control characters
More comparisons against an alternat implementation showed this case

```
[A]:\u{1b}

X [A] Y
```

In this case the `[A]:\u{1b}` is being parsed as a link-defintion, giving: 

```
<p>X <a href=\"%1B\">A</a> Y</p>
```

but the specs, both 0.30 and 0.29-gfm, say that a link destination is (emph mine)

> ...
>  * a nonempty sequence of characters that does not start with <, *does not include ASCII space or control characters*, and includes parentheses only if (a) they are backslash-escaped or (b) they are part of a balanced pair of unescaped parentheses. (Implementations may impose limits on parentheses nesting to avoid performance issues, but at least three levels of nesting should be supported.)

So this should parse as:

```
<p>[A]:\u{1b}</p>
<p>X [A] Y</p>
```

mikeando:
link destinations are not supposed to include ascii control characters - fixes #218

### DIFF
--- a/src/parser/inlines.rs
+++ b/src/parser/inlines.rs
@@ -1152,7 +1152,7 @@ pub fn manual_scan_link_url_2(input: &[u8]) -> Option<(&[u8], usize)> {
             }
             nb_p -= 1;
             i += 1;
-        } else if isspace(input[i]) {
+        } else if isspace(input[i]) || input[i].is_ascii_control() {
             if i == 0 {
                 return None;
             }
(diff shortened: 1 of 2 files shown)


kivikakk:
Thank you! :heart:

kivikakk:
(I've slightly modified your PR body (`see #218` -> `fixes #218`) so that merging this PR closes the relevant issue.

PR #222, opened 2022-06-08
Add support for WASM compilation

benwis:
I wanted to use comrak on the web, and I needed a couple small tweaks to compile to WASM.

Mostly I needed to modify the dependencies to reduce the number of features for syntect, as regex-onig is not portable. Also, there's no way to have a default config file in WASM.

I am not using most of the features of comrak, so I am really not sure what depends on syntect's features, but compiling markdown to syntax highlighted HTML is working with these changes.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -22,6 +22,20 @@ jobs:
       run: env SPEC=false script/cibuild
     - name: Run spec tests
       run: env SPEC=true script/cibuild
+  build_wasm:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+      with:
+        submodules: true
+    - name: Obtain Rust
+      run: rustup override set stable
+    - name: Setup for wasm
+      run: rustup target add wasm32-unknown-unknown
+    - name: Build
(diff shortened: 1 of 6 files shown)


kivikakk:
Hey there; thanks so much for this! I'm happy to accept these changes, but I'd like a CI job that compiles to wasm like you do so I can be sure not to break this in the future by accident.

I'm super unfamiliar, and my attempt I've pushed to your PR branch in f06c360 doesn't actually compile — see [the build log](https://github.com/kivikakk/comrak/runs/6841331870?check_suite_focus=true). Would you be able to adjust it so it reflects how you do this? Thanks! :blush:

benwis:
Ah, there's nothing wrong with your action. But you need Cargo's better cfg feature resolver to actually stop it from trying to import onig in this condition. Alternatively you could update to the 2021 edition

kivikakk:
Finally got there! Let me know if you want a release or anything.

benwis:
Would love a release, thanks for finding the issue. Guess propfuzz doesn't like it either

SpanishPear:
Would also love a release for this!

kivikakk:
These are now available in [0.13.1](https://github.com/kivikakk/comrak/releases/tag/0.13.1).

PR #228, opened 2022-07-06
Fix wrapping on lines succeeded by unwrappable lines

### LINKED ISSUE #227: Lines wrap incorrectly when followed by a non-wrappable line
Hello! I believe I've found a bug. I have a markdown file `untitled.md`:

```md
item that doesn't need wrapping

a-long-line-that-won't-be-wrapped-because-there-is-no-character-it-can-break-on
```

When running this through the cli with the `--to commonmark` and `--width 72` options:

```sh
comrak --to commonmark --width 72 untitled.md
```

It outputs the following:

```md
item that doesn't need
wrapping

a-long-line-that-won't-be-wrapped-because-there-is-no-character-it-can-break-on
```

I think the list item is not supposed to wrap there. If the last line is short enough to not need wrapping, this doesn't occur. If the last line is wrappable (e.g. remove all the `-` characters), this also doesn't occur.

edwardloveall:
When outputting as `commonmark` lines followed by a long, unwrappable
section would wrap incorrectly. For example:

```md
item that doesn't need wrapping

a-long-line-that-won't-be-wrapped-because-there-is-no-character-it-can-break-on
```

When running this with the `--to commonmark` and `--width 72` options:

```sh
comrak --to commonmark --width 72 some-markdown.md
```

It outputs the following:

```md
item that doesn't need
wrapping

a-long-line-that-won't-be-wrapped-because-there-is-no-character-it-can-break-on
```

What was happening is that when a new line is needed, the algorithm
resets a bunch of variables like the current column, if we're at the
beginning of a line, etc. One of those variables is the position of the
last place we can break a line: the last non-space character. This is
stored in `CommonMarkFormatter`s `last_breakable` field.

cmark-gfm [resets this][cmark-gfm-render] in `render.c`, and now so does comrak.

This also adds a new `commonmark` test type that doesn't write to HTML
to test commonmark specific features or bugs.

[cmark-gfm-render]: https://github.com/github/cmark-gfm/blob/f26f75ce5dbfeddf8c3e3df10a6f238f65f8d6ba/src/render.c#L60

Fixes #227

### DIFF
--- a/src/cm.rs
+++ b/src/cm.rs
@@ -108,6 +108,7 @@ impl<'a, 'o> CommonMarkFormatter<'a, 'o> {
                 }
             }
             self.column = 0;
+            self.last_breakable = 0;
             self.begin_line = true;
             self.begin_content = true;
             self.need_cr -= 1;
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -62,6 +62,18 @@ fn compare_strs(output: &str, expected: &str, kind: &str) {
 }
 
 #[track_caller]
+fn commonmark(input: &str, expected: &str) {
+    let arena = ::Arena::new();
+    let mut options = ::ComrakOptions::default();
+    options.render.width = 72;
+
+    let root = ::parse_document(&arena, input, &options);
+    let mut output = vec![];
+    cm::format_document(root, &options, &mut output).unwrap();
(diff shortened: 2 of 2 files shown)


kivikakk:
Wonderful — thank you so much!

PR #236, opened 2022-07-17
Demonstrate that strikethrough works erroneously, maybe

gjtorikian:
The GFM spec [states](https://github.github.com/gfm/#strikethrough-extension-) that

> Strikethrough text is any text wrapped in two tildes (~).

But as I show in this failing text, a single tilde is enough to trigger a strikethrough. 

_But the plot thickens!_ Because in a GitHub comment box, a single tilde also triggers a strikethrough:

```
Hello **world** -- how are _you_ today? I'm ~~fine~~, ~yourself~?
```

Hello **world** -- how are _you_ today? I'm ~~fine~~, ~yourself~?

* * *

I'm not sure who is correct and who isn't here.

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -635,11 +635,13 @@ fn strikethrough() {
         concat!(
             "This is ~strikethrough~.\n",
             "\n",
-            "As is ~~this, okay~~?\n"
+            "As is ~~this, okay~~?\n",
+            "But not ~this~.\n"
         ),
         concat!(
             "<p>This is <del>strikethrough</del>.</p>\n",
-            "<p>As is <del>this, okay</del>?</p>\n"
+            "<p>As is <del>this, okay</del>?</p>\n",
+            "<p>But not this.</p>\n"
         ),
     );
 }


kivikakk:
This vaguely rang a bell … https://github.com/github/cmark-gfm/issues/71#issuecomment-355855609

So the GFM spec is wrong, for sure.

gjtorikian:
Oh no, what hornet's next have I walked into...

Is the _spec_ wrong, or is _GitHub's implementation of its own spec_ wrong? It seems to me the latter. Regardless, I would like to propose that this lib respect the two tilde behavior to match the spec, rather than the one tilde default or the existence of some kind of config (a la `CMARK_OPT_STRIKETHROUGH_DOUBLE_TILDE`). If that sounds okay, I can do the work to implement it in Rust. Either that, or (dun dun dun) comrak has its own spec (it already has [its own extensions](https://github.com/kivikakk/comrak#extensions), and I've already been thinking about finding the time to document them).

At the end of the day I personally don't care what the behavior is as long as it's documented and unambiguous.

kivikakk:
> Is the _spec_ wrong, or is _GitHub's implementation of its own spec_ wrong?

That's a great question, and one I wish I could ask the spec author.

Thankfully, I'm very practiced at talking to myself, so this is a piece of cake!

The spec is wrong. When I wrote it during the Sundown→CommonMark transition, we (i.e. Vicent and I) were basically under the belief that strikethrough using a single tilde was extremely rare, and so I had the spec accommodate that belief, but didn't actually reflect on the behaviour of the strikethrough extension as implemented.

It turned out to be extremely not the case that single tildes were rare — iirc https://github.com/github/cmark-gfm/issues/71#issuecomment-355855609 was written after actually analysing the entire comment table. I **should** have updated the spec at that instant. The fact that the spec doesn't have any counterexamples (i.e. one, three tildes) is another example of my quality™ writing.

I'll investigate if there's someone @github I can ask nicely to see if they wanna bump the gfm spec repo. 😅 

Either way, I would like the extensions of comrak that correspond to GFM extensions to match their counterparts quite closely when it comes to behaviour not covered by the spec. Seeing as the spec _doesn't_ contain counterexamples, comrak technically remains spec-compliant while supporting single tildes in much the same way the reference implementation does.

I know it's not a lovely situation in many respects, but I definitely broke a few eggs making that omelette.

gjtorikian:
Sounds good to me! I'm sorry for opening this can of worms. 😅 Your explanation makes complete sense, and in fact, I like that `~` operates much the same way as `*` or `_`. Too bad we can't use `^H` as a glyph for this. 🤓

PR #237, opened 2022-07-27
Allow `c_char` options to be NULL

gjtorikian:
What the title says! 

`ComrakOptions::default()` sets the `header_ids` option of `extensions` to `""`; this means that, by default, headers are always given anchor tags (with an ID matching the text content of the header).

In C, there's currently no way of implementing the removal of the `header_ids` option. This won't work:

```c
comrak_set_extension_option_header_ids(comrak_options, "", 0);
```

Because it reads just as `default` does: add the header IDs, but don't add any prefix.

This PR introduces (and tests) a different possibility:

```c
comrak_set_extension_option_header_ids(comrak_options, NULL, 0);
```

Passing this ☝️ will effectively "turn off" header ID generation by checking to see if the incoming value is `NULL`. (Some kind of `NULL` check is necessary here, with or without this change. I was able to segfault the FFI.)

I'm happy to expose a different function to configure this if it doesn't look good!

### DIFF
--- a/c-api/src/comrak_options.rs
+++ b/c-api/src/comrak_options.rs
@@ -31,9 +31,14 @@ macro_rules! make_c_char_option_func {
                 v_len: size_t,
             ) {
                 let comrak_options = to_ref_mut!(c_comrak_options);
-                let value = unwrap_or_ret_err_code! { to_str!(v, v_len) };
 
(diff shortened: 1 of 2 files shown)


kivikakk:
This is just perfect, thanks! We might want to be able to turn the other optional string options back off, too, and this allows for that nicely.

PR #238, opened 2022-08-03
Replace `lazy_static` dependency with `once_cell`

Turbo87:
https://github.com/tokio-rs/tokio/pull/3187 explains the rationale quite well. `once_cell` provides the same functionality, but without relying on macros :)


kivikakk:
Thank you so much! Much nicer.

PR #239, opened 2022-08-09
Adding functionality to build SyntectAdapters with custom themes, syntax sets, etc.

ArvinSKushwaha:
I came across the issue of being unable to add custom themes to SyntectAdapters, so to avoid making the internals of SyntectAdapters public, I made a small builder to let users build custom SyntectAdapters.

Thank you for your time!


kivikakk:
Thank you so much for this! The CI error is my issue to deal with; I'll get that cleaned up and merge when done!

PR #242, opened 2022-09-13
Make `comrak --help` readable on my terminal

mgeisler:
When I run `comrak --help` on my terminal, I see unwrapped and very hard-to-read output.

This is what I see:

image

The output is similar with a 100 columns terminal:

image

With the `wrap_help` Cargo option for Clap, we can make it produce much more readable output. The output will even adapt to the terminal width (similar to man(1)) and thus look good in many situations:

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -150,6 +150,7 @@ dependencies = [
  "atty",
  "bitflags",
  "strsim",
+ "term_size",
  "textwrap",
  "unicode-width",
  "vec_map",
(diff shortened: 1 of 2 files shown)


mgeisler:
When playing with this, I noticed that the text finally wrapped correct with a terminal which is 120 columns wide. I'm basically suggesting not to hard-code this and instead use a millisecond or two at runtime to adapt the text to match :-)

kivikakk:
This is a wonderful contribution — thanks so much, really appreciate it!

mgeisler:
Hi @kivikakk, thanks, I'm very happy to help where I can!

PR #248, opened 2022-11-13
Bump clap version to 4.0 and switch to the Derive API

tranzystorek-io:
Hi, hope this isn't too big for a first contribution 🙂

Should make the code a bit more readable with the Derive API.

Notes:

- had to bump pest in `Cargo.lock` to resolve a dependency conflict
- the extension flag now additionally accepts comma-delimited lists, e.g. `-e strikethrough,table,footnotes`
- clap no longer adds author info by default, so I skipped it for now (adding it back is a bit involved)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,15 +18,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ansi_term"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


tranzystorek-io:
Seems we should somehow disable clap for wasm: https://github.com/lambda-fairy/rust-errno/issues/21#issuecomment-608122469

kivikakk:
Thank you so much, this is perfect, and certainly not too big at all! Really clear changes and very easy to review. Massively appreciated! 🥰

PR #249, opened 2022-11-14
c-api: remove

kivikakk:
Seeing as commonmarker doesn't use the C API to do FFI with comrak any more, I figured I'd remove it — I don't want to maintain it without an actual user, especially given it's C and sharp edges can hurt when forgotten about.

Just wanted to run this by you, @gjtorikian, in case you had other thoughts!


gjtorikian:
I completely agree! C API, we hardly knew ye.

PR #250, opened 2022-11-18
Add option to not have aria-hidden attribute on header links

Basicprogrammer10:
Adds a `header_no_aria_hidden` option on the `ComrakExtensionOptions` struct to allow for header ID links which don't have the aria-hidden attribute.


kivikakk:
Hi there! Thanks for the PR :) I'm not opposed to this, but I have a few things that came to mind:

1. The option should probably be called `header_ids_[…]` as it only applies when the `header_ids` option is turned on.
2. On the topic of naming, I think we're better off having the argument called `header_ids_aria_hidden` and defaulting to `true`, rather than `header_ids_no_aria_hidden` and `true` turning `aria-hidden` off. Especially because the whole attribute as-added is `aria-hidden="true"`, something about specifying `--header-ids-no-aria-hidden=true` to not show `aria-hidden="true"` and `--header-ids-no-aria-hidden=false` to show `aria-hidden="true"` just does my head in.
3. Could you teach me a little bit on what the intention is in removing this attribute? The header ID extension only exists to emulate GitHub's header IDs, which include `aria-hidden="true"`. I'm not super across ARIA so it's not clear to me if this is in fact inappropriate in general.

kivikakk:
I'm going to close this PR due to the lack of response; in general I think such tweaks are probably better suited by post-processing. Otherwise we'll end up with a thousand options for toggling every last bit of output.

PR #252, opened 2022-12-03
Track which symbol was used to mark task item as checked

felipesere:


src/parser/mod.rs
```diff
@@ -1676,7 +1676,13 @@ impl<'a, 'o, 'c> Parser<'a, 'o, 'c> {
         }
 
         *text = text[end..].to_vec();
-        let checkbox = inlines::make_inline(self.arena, NodeValue::TaskItem(active));
+        let checkbox = inlines::make_inline(
+            self.arena,
+            NodeValue::TaskItem {
+                checked: symbol != b" ",
+                symbol: symbol[0] as char,
```
kivikakk, line 1683: Because we're only matching a single byte in the regex, I think we are better off leaving this field as `u8` rather than casting it to `char` (which could imply a full set of Unicode characters might be stored here). This keeps us consistent with the other stored characters, `bullet_char` and `fence_char`.

src/tests.rs
```diff
@@ -1389,7 +1391,7 @@ fn exercise_full_api<'a>() {
         ::nodes::NodeValue::Text(text) => {
             let _: &Vec<u8> = text;
         }
-        ::nodes::NodeValue::TaskItem(checked) => {
+        ::nodes::NodeValue::TaskItem { checked, .. } => {
             let _: &bool = checked;
```
kivikakk, line 1395: Since we're adding to the API, let's add a similar "test" for the `symbol` field here.

src/cm.rs
```diff
@@ -328,7 +328,7 @@ impl<'a, 'o> CommonMarkFormatter<'a, 'o> {
             NodeValue::HtmlInline(ref literal) => self.format_html_inline(literal, entering),
             NodeValue::Strong => self.format_strong(),
             NodeValue::Emph => self.format_emph(node),
-            NodeValue::TaskItem(checked) => self.format_task_item(checked, entering),
+            NodeValue::TaskItem { checked, .. } => self.format_task_item(checked, entering),
```
kivikakk, line 331: In order to roundtrip these alternate character types when outputting CommonMark, we'll need to pass `symbol` through to `format_task_item` and use it there.

kivikakk (changes requested):
This is looking excellent — thank you! I have a few comments on cleaning this up for merge which I think should be no trouble at all.

The only blocking issue is the now-failing spec tests. The [GFM spec for task list items](https://github.github.com/gfm/#task-list-items-extension-) only recognises task list items where the marker contains whitespace, `x` or `X` in the square brackets. We're now recognising tasklists with anything inside those square brackets.

Because we're not only saving additional information about the input, but also changing the interpretation of it, we'll need to use the relaxed regex only when an flag/option is specified.

src/parser/mod.rs
felipesere: Considering if we should rename it to `task_char`?

kivikakk: We certainly could! I don't mind either way.

kivikakk:
Great! As soon as we have the new option, we're good to go. Did you want to do that, or would you prefer I did?

felipesere:
I totally forgot about the flag 😅

kivikakk:
Awesome, thank you so much! Free-standing is totally fine for the option.

I've added two commits — one (a5efb05) to demonstrate an issue, and one (59e2ef5) with a suggested fix. The problem is that the regex is only initialised once for the lifetime of the program, so the first time a tasklist is processed, the setting of the option at that time is then used for all subsequent invocations.

If you're happy enough with my fix, let me know, but I didn't want to dictate the answer seeing as it's your feature! Feel free to amend it or drop the commit and do it a different way if you want. :)

felipesere:
Yeah this looks brilliant 👍

kivikakk:
Awesome. Thank you so much!

PR #256, opened 2022-12-06
improve tagfilter performance

### LINKED ISSUE #255: `tagfilter` can be a bottleneck
When converting markdown with many HTML tags `tagfilter` can become the bottleneck. This is because we convert bytes to utf8 and lowercase them in a loop, see [here](https://github.com/kivikakk/comrak/blob/b0b29f34b7d23255eedb2cd67ffc040f15b5580a/src/html.rs#L198).

fiji-flo:
Within tagfilter we currently run the same line for every iteration. Let's move this out of the loop.

fixes #255

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -194,11 +194,9 @@ fn tagfilter(literal: &[u8]) -> bool {
         i += 1;
     }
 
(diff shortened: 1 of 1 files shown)


kivikakk:
Thank you!

PR #260, opened 2022-12-22
[ShortCode] Add support for gemojis via shortcodes extension

eklipse2k8:
This adds support for shortcode extension to render gemoji's.


gjtorikian:
This is an incredibly well-encomposing PR!

kivikakk:
This is beautiful — thank you so much! Let me know if you'd like a new release for this.

gjtorikian:
> Let me know if you'd like a new release for this.

I personally would, but I can always pin to a SHA.

If it's helpful, I recently wrote a GitHub Workflow that tags a new GH Release, adds a changelog, and deploys to crates: https://github.com/gjtorikian/escapist/blob/main/.github/workflows/tag_and_release.yml


It's triggered by `workflow_dispatch`—I haven't quite figured out how to fire it only when the version number in Cargo.toml changes—but that's good enough!

Issue #259, opened 2022-12-21
Generated footnote html different from cmark-gfm


digitalmoksha:
I've started doing some testing between `comrak` and `cmark-gfm`.

I've noticed that the footnote extension generates different html. In particular, `data-footnote-ref`, `data-footnotes`, `data-footnote-backref`, and `aria-label` attributes are no longer added.

**input**

```markdown
1. one[^1]
[^1]: AbC
```

**cmark-gfm output**

```html
<ol>
<li>one<sup class="footnote-ref"><a href="#fn-1" id="fnref-1" data-footnote-ref>1</a></sup></li>
</ol>
<section class="footnotes" data-footnotes>
<ol>
<li id="fn-1">
<p>AbC <a href="#fnref-1" class="footnote-backref" data-footnote-backref aria-label="Back to content">↩</a></p>
</li>
</ol>
</section>
```

**comrak output**

```html
<ol>
<li>one<sup class="footnote-ref"><a href="#fn1" id="fnref1">1</a></sup></li>
</ol>
<section class="footnotes">
<ol>
<li id="fn1">
<p>AbC <a href="#fnref1" class="footnote-backref">↩</a></p>
</li>
</ol>
</section>
```

I also notice the footnote ids no longer use a `-`, such as `fn-1` vs `fn1`

Are these changes by design, or missing implementation?

digitalmoksha:
If it's missing implementation I can put together a PR, but if it's by design I won't worry about it.

kivikakk:
Not by design — when first implemented it was 1:1, but it looks like GitHub have since added these attributes: https://github.com/github/cmark-gfm/pull/234

Happy to accept PRs that realign.

kivikakk:
With the merge of #273, we're a lot closer. There's a few smaller things:

* As @froh mentioned in #271, the footnote label content can differ. We should align this.
* As noted at https://github.com/kivikakk/comrak/pull/273#issuecomment-1403056774, there are some other differences in the IDs generated and the acceptable characters in labels. These should also be aligned. (This is currently causing spec failure in CI.)

digitalmoksha:
Currently `comrak` sequentially numbers footnotes, while `cmark-gfm` utilizes the footnote name.  For example:

```
first[^1] and second[^😄second] and twenty[^_twenty]

[^1]: one
[^😄second]: two
[^_twenty]: twenty
```

in `comrak`

```
[...]

PR #273, opened 2023-01-20
Add footnote attributes that mirror cmark-gfm

digitalmoksha:
This aligns the footnote output with `cmark-gfm`. Relates to https://github.com/kivikakk/comrak/issues/259

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 
 jobs:
-  build_test_spec:
+  build_test:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


kivikakk:
Hi @digitalmoshka — thanks so much for this, this is great.

The spec builds are failing, but that's only because the `extensions.txt` spec in our vendored `cmark-gfm` doesn't match the new output.  I tried pulling in the current upstream for that part of the file, and it also doesn't quite work due to some remaining differences in footnote ID generation (and a parsing issue with footnotes like `` Hello[^"><script>alert(1)</script>] ``, which are apparently supported!).

That's fine with me — I'll address it in a follow-up PR. This here is great.

digitalmoksha:
Thanks @kivikakk for getting this over the line 🚀

PR #285, opened 2023-03-26
Change em and strong to i and b

snan:
Normally Markdown's default of using em for emphasis makes sense.
Emphasis is what you use italics for most of the time. When you need
cite, you can type <cite> and when you need a more generic italic, for
example for foreign language words or Linnaean names, you can type
<i>. But on user-writable forums that disable fallthrough HTML, i is
the better default. It's never "wrong", per se (em is just more
specific) and it matches how many people sometimes use * and _ on
these forums, when they think "I want to make this look cursive"
rather than "I need to semantically emphasize this".

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ And there's a Rust interface. You can use `comrak::markdown_to_html` directly:
 ``` rust
 use comrak::{markdown_to_html, ComrakOptions};
 assert_eq!(markdown_to_html("Hello, **世界**!", &ComrakOptions::default()),
-           "<p>Hello, <strong>世界</strong>!</p>\n");
+           "<p>Hello, <b>世界</b>!</p>\n");
 ```
 
 Or you can parse the input into an AST yourself, manipulate it, and then use your desired formatter:
--- a/examples/sample.rs
+++ b/examples/sample.rs
@@ -5,7 +5,7 @@ fn small() {
 
     assert_eq!(
         markdown_to_html("Hello, **世界**!", &ComrakOptions::default()),
-        "<p>Hello, <strong>世界</strong>!</p>\n"
+        "<p>Hello, <b>世界</b>!</p>\n"
     );
 }
 
--- a/src/adapters.rs
+++ b/src/adapters.rs
@@ -32,7 +32,7 @@ pub struct HeadingMeta {
     pub level: u8,
 
     /// The content of the heading as a "flattened" string&mdash;flattened in the sense that any
-    /// `<strong>` or other tags are removed. In the Markdown heading `## This is **bold**`, for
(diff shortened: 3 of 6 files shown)


kivikakk:
Hi there! Unfortunately this change would break CommonMark compatibility, so it's not one I can accept. Comrak is first and foremost a CommonMark library, and spec compliance takes priority.

Additionally, I'm a little bit confused by the rationale -- I'm not sure why one would convert Markdown to HTML only to then feed that HTML into a forum's user interface.

snan:
> I'm not sure why one would convert Markdown to HTML only to then feed that HTML into a forum's user interface.

Ah, no, that's not what I was saying at all. What I meant was that forum user interfaces use Comrak in safe mode (which is appropriate), people type in `*` and `_` in there, and they get the wrong thing, which is em when they meant i.

With CommonMark, it's possible to write:

    *this* to get <em>this</em>
    <cite>this</cite> to get <cite>this</cite>
    <i>this</i> to get <i>this</i>

`<i>` is the more general one. It is a superset of the other ones. People choose `<em>` or `<cite>` only when they want to be more specific.

Emphasis is for when you `<em>`really`</em>` mean something, cite is for when you mention name of a book, like `<cite>`Structure and Interpretation of Computer Programs`</cite>`, and the plain old italics element can handle those cases as well as cases not covered by them, like names of trees like `<i>`Gingko biloba`</i>`.

When it's possible to write all three, which CommonMark can (by being able to enter HTML directly), it's fine that the "shortcut" (using `*` or `_`) is for the most common one, which is em.

When it is not possible to write all three, and Comrak cannot (unless people turn on `--unsafe`, which, for web forums is a bad idea), the "shortcut" (using `*` or `_`) should emit the most general one, which is i. `<i>`This`</i>` is not wrong, but this is: `<em>`Gingko biloba`</em>`. 

Comrak already is incomaptible with CommonMark unless you use `--unsafe`. (If you wanna make the "shortcut" (using `*` or `_`) emit em in `--unsafe` mode and i in safe mode, that'd be fine by me.)

According to [when I brought this up over at CommonMark](https://github.com/commonmark/commonmark-spec/issues/652) a few years ago, they wrote:

> Implementers (or plugin authors) could decide to use asterisk `*` for `em` / `strong` and underscore `_` for `i` / `b` or vice versa

kivikakk:
Alright, I see what you mean!

I'm not inclined to change the default and change the output from that suggested by the spec (e.g. [all of these](https://spec.commonmark.org/0.30/#example-350)), surprised downstream users, etc. The majority of implementations will follow the suggested output in the spec, given its [double status as a conformance test](https://spec.commonmark.org/0.30/#about-this-document).

But I'm happy to take a PR which introduces an option or options to switch these in the output.

Issue #291, opened 2023-03-29
A better mechanism for extending, toggling, option-ing, adapting, adjusting …


kivikakk:
It's become clear to me that Comrak's options and formatting systems are struggling under the weight of what we'd all like it to do slightly differently. Some examples:

* AST
  * Add new node types.
* Parsing
  * Disable built-in parsing rules, block and inline.
  * Add new parsing rules, block and inline.
  * Add new post-processing rules.
* Formatting
  * Include (or omit) certain attributes in HTML tags for certain node types.
  * Completely change how certain node types are formatted, while retaining the built-in behaviour for others.

I'd like to add to this list to get an idea of where extensibility is actually called for, in order to work out how to provide it in Comrak's internal design with more orthogonality than the current "add another field to an options struct and sprinkle conditionals everywhere" design. Maintaining performance is a key consideration.

If anyone has anything to add to this list, or design suggestions for ways to achieve this goal, please comment!

Refs: #268, #261, #259, #258, #253, #251, #246, #244, #208, #204, #180, #164, #149, #137, #136, #133, #132, #130, #129, #52, #48, #40, #23, etc.

YJDoc2:
Hey, may I post a tldr-type comment that outline each of the issue/pr you have mentioned, so to have a better overview of all stuff? Or would it be too noisy to have that giant-ish comment?

kivikakk:
If you feel like that'd help, by all means feel free! I think a giant-ish comment is totally fine.

PR #292, opened 2023-03-29
Add builder derive and non_exhaustive for option structs

YJDoc2:
This implements the suggestions from https://github.com/kivikakk/comrak/pull/174#issuecomment-950273881

This adds builder for each of the option structs and a `non_exhaustive` annotations on all of them.

It goes without saying that this is a breaking change due to the `non_exhaustive`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -130,6 +130,7 @@ version = "0.18.0"
 dependencies = [
  "arbitrary",
  "clap",
+ "derive_builder",
  "emojis",
  "entities",
  "memchr",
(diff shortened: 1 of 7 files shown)


kivikakk:
Hmmm, yep, I totally get what you mean -- it does seem pretty verbose. I think it does make sense to remove the distinction between the three different structs when using builders; that was essentially to make it easier to provide a smaller set of options, which builders help with.

That said, I'm a bit worried that it'll be less obvious what distinct kinds of options are available. This definitely ties into #291 in a big way, so I'm not exactly sure yet which direction to go.

Thanks so much for this -- I'm going to think on it now that I can really see what it looks like. I think it's close but not quite it.

kivikakk:
Thanks for your work here, I really appreciate it! I've rebased it after this came up again in #303, #305 and #321 (!). One thing I left out from that (and this) PR is the root options struct. I'm still not convinced that we shouldn't just merge all the options into one big struct, but it still seems too ugly -_-

YJDoc2:
Glad to help!

Turbo87:
```
error[E0433]: failed to resolve: could not find `OptionsBuilder` in `comrak`
  --> crates_io_markdown/lib.rs:64:17
   |
64 |         comrak::OptionsBuilder::default();
   |                 ^^^^^^^^^^^^^^ could not find `OptionsBuilder` in `comrak`
```

maybe I'm using it wrong, but it looks like those builder structs are not `pub`. maybe introduce a `fn builder(Self)` on the options structs?

PR #295, opened 2023-03-29
Add quadratic fuzz target

philipturnbull:
This is the fuzzer which discovered the issues fixed by GHSA-8hqf-xjwp-p67v. The comment at the top of `quadratic.rs` explains how the fuzzer works. I ran the fuzzer for ~1 hour on top of the latest `main` branch and it didn't trigger any quadratic parsing issues. Previously, it would trigger a quadratic parsing bug in a few minutes, so the parser seems pretty solid 🎉

The fuzzer can be run with `cargo fuzz run quadratic`

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -42,3 +42,9 @@ name = "gfm"
 path = "fuzz_targets/gfm.rs"
 test = false
 doc = false
+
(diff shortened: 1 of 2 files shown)


kivikakk:
Thanks so much, this is awesome! 🤍

PR #297, opened 2023-04-11
Simplify anchorize()

kornelski:
Fewer string copies.

This function could be simplified further, but I'm surprised that it's not using the `slug` crate, which is already a dependency of `comrak`.

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -105,21 +105,19 @@ impl Anchorizer {
         static REJECTED_CHARS: Lazy<Regex> =
             Lazy::new(|| Regex::new(r"[^\p{L}\p{M}\p{N}\p{Pc} -]").unwrap());
 
-        let mut id = header;
-        id = id.to_lowercase();
-        id = REJECTED_CHARS.replace_all(&id, "").to_string();
-        id = id.replace(' ', "-");
+        let mut id = header.to_lowercase();
+        id = REJECTED_CHARS.replace_all(&id, "").replace(' ', "-");
 
         let mut uniq = 0;
         id = loop {
             let anchor = if uniq == 0 {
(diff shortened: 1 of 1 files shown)


kivikakk:
> This function could be simplified further, but I'm surprised that it's not using the `slug` crate, which is already a dependency of `comrak`.

I almost didn't notice that happen! It crept in just for an example, recently (fc6502b). `Anchorizer` mimics the exact output of GitHub's own slug generation, which is why it doesn't use any crate.

kivikakk:
Thank you! 🤍

vadimkantorov:
@kivikakk Does GitHub actually use this `anchorize()` in production to slugify? or is it a somewhat outdated reverse-engineered attempt?

kivikakk:
@vadimkantorov at the time of writing, it matched the actual implementation behaviour exactly (which I also maintained), but Comrak wasn't used by GitHub.

Whether it is now -- they've invested some developer time and sponsorship in it, but that's all I know -- or whether it still matches, I couldn't say.

vadimkantorov:
Interesting, thank you!

https://github.com/github/markup?tab=readme-ov-file#markups refers to https://github.com/gjtorikian/commonmarker which is a ruby wrapper of https://github.com/kivikakk/comrak which is a rust port of https://github.com/github/cmark-gfm/

So I wondered if this slug generation lives in comrak or is also implemented in cmark-gfm?

gjtorikian:
Feel free to contact GitHub Support, they would know best about what they are currently using in production: http://github.com/support

vadimkantorov:
Sure, thanks for the tip, in the meanwhile I also asked at https://github.com/orgs/community/discussions/114288, will contact the support as well

PR #302, opened 2023-05-06
Adding support for subscript, highlighted and underlined text

blood-rogue:
- One '~' produces a subscript instead of strikethrough (when enabled)
- "==" highlights the text inside it
- "++" underlines the text inside


kivikakk:
Hi there! Thanks so much for this high-quality submission!

There's one major issue, which is that this changes the behaviour of the strikethrough extension, even when subscript isn't enabled. This causes an incompatibility with [GitHub Flavored Markdown](https://github.github.com/gfm/#strikethrough-extension-).

I'd be happy to accept the proposed change if this was addressed -- I'd be okay with subscript being enabled changing strikethrough's behaviour accordingly, but not with it disabled (i.e. GFM defaults). Right now there's an "interesting" bug that results where enabling only strikethrough (not subscript) and then using one `~` results in emphasis being rendered (!); see https://github.com/kivikakk/comrak/actions/runs/4902153007/jobs/8760508796?pr=302#step:4:659.

I'll leave some more comments inline.

src/main.rs
```diff
@@ -209,6 +211,8 @@ fn main() -> Result<(), Box<dyn Error>> {
             front_matter_delimiter: cli.front_matter_delimiter,
             #[cfg(feature = "shortcodes")]
             shortcodes: cli.gemojis,
+            subscript: exts.contains(&Extension::Subscript),
+            highlight: exts.contains(&Extension::Highlight)
```
kivikakk, line 215: `cargo fmt` will add a trailing `,` here.

src/nodes.rs
```diff
@@ -135,7 +135,7 @@ pub enum NodeValue {
     /// per the GFM spec.
     Strikethrough,
 
-    /// **Inline**.  Superscript.  Enabled with `ext_superscript` option.
+    /// **Inline**.  Superscript.
```
kivikakk, line 138: Good catch, thanks!

src/nodes.rs
```diff
@@ -151,6 +151,12 @@ pub enum NodeValue {
     #[cfg(feature = "shortcodes")]
     /// **Inline**. An Emoji character generated from a shortcode. Enable with feature "shortcodes".
     ShortCode(NodeShortCode),
+
+    /// **Inline**.  Supscript.
```
kivikakk, line 155:
```suggestion
    /// **Inline**.  Subscript.
```

src/parser/inlines.rs
```diff
@@ -337,10 +343,11 @@ impl<'a, 'r, 'o, 'd, 'i, 'c, 'subj> Subject<'a, 'r, 'o, 'd, 'i, 'c, 'subj> {
...
                     b'^' => 1,
-                    b'"' => 2,
-                    b'\'' => 3,
-                    b'_' => 4,
-                    b'*' => 5 + (if c.can_open { 3 } else { 0 }) + (c.length % 3),
+                    b'=' => 2,
+                    b'"' => 3,
+                    b'\'' => 4,
+                    b'_' => 5,
+                    b'*' => 6 + (if c.can_open { 2 } else { 0 }) + (c.length % 3),
```
kivikakk, line 350: I think we don't want to change this `3` to `2` here -- it causes an overlap in the `c.can_open` and `!c.can_open` cases. (Before, we had effectively `0+(c.length%3)` -- 0, 1, 2 -- and `3+(c.length%3)` -- 3, 4, 5. Now it looks like 0, 1, 2, and 2, 3, 4.)

Instead, we should keep this like this:

```suggestion
                    b'*' => 6 + (if c.can_open { 3 } else { 0 }) + (c.length % 3),
```

and add an extra element in the array here, by changing `11` to `12`:

https://github.com/kivikakk/comrak/blob/bb6ff755ba850e70b6477fc935a94f4da272dc57/src/parser/inlines.rs#L320

src/parser/inlines.rs
```diff
@@ -837,13 +845,6 @@ impl<'a, 'r, 'o, 'd, 'i, 'c, 'subj> Subject<'a, 'r, 'o, 'd, 'i, 'c, 'subj> {
         opener_num_chars -= use_delims;
         closer_num_chars -= use_delims;
 
-        if self.options.extension.strikethrough
-            && opener_char == b'~'
-            && (opener_num_chars != closer_num_chars || opener_num_chars > 0)
```
kivikakk, line 842: I suspect (though haven't checked) this will cause spec failure with [this example](https://github.github.com/gfm/#example-493). The spec requires three or more tildes to not generate a strikethrough; since we've already subtracted `use_delims` (2 for a strikethrough of length 2 or more) from `opener_num_chars`, if we had 3 or more, `opener_num_chars > 0` would be true and so we'd `return None` here.

src/parser/mod.rs
```diff
@@ -343,6 +343,28 @@ pub struct ComrakExtensionOptions {
...
     pub shortcodes: bool,
+
+    /// Enables the subscript Comrak extension.
+    ///
+    /// ```
+    /// # use comrak::{markdown_to_html, ComrakOptions};
+    /// let mut options = ComrakOptions::default();
+    /// options.extension.subscript = true;
+    /// assert_eq!(markdown_to_html("Water is H~2~O.\n", &options),
+    ///            "<p><p>Water is H<sup>2</sup>O.</p></p>\n");
```
kivikakk, line 354:
```suggestion
    ///            "<p>Water is H<sup>2</sup>O.</p>\n");
```

src/tests/strikethrough.rs
```diff
@@ -5,7 +5,7 @@ fn strikethrough() {
     html_opts!(
         [extension.strikethrough],
         concat!(
-            "This is ~strikethrough~.\n",
+            "This is ~~strikethrough~~.\n",
```
kivikakk, line 8: This behaviour (when _only_ `strikethrough` is enabled, and not `subscript` too) should be unchanged, and so we should leave the test as-is.

blood-rogue: Will try to resolve strikethrough and subscript conflicts.

src/parser/inlines.rs
blood-rogue: If subscript is enabled, it will be a subscript _as well as_ a strikethrough (if both enabled) like in cases of italics and bolds. Considering this I removed the condition.

blood-rogue:
Can I add a new feature of `inserted` text i.e. basically underline using `++`? Got inspiration from https://markdown-it.github.io/

src/parser/inlines.rs
blood-rogue: Adding and additional `&&` to check if subscript is enabled. Like
```rust
if (self.options.extension.strikethrough
    && opener_char == b'~'
    && (opener_num_chars != closer_num_chars || opener_num_chars > 0))
    && !self.options.extension.subscript
{
    return None;
}
```

blood-rogue:
Can I get some info regarding the `skip_chars` array in `Subject` struct in `src/parser/inline.rs@41:4`?

blood-rogue:
Any other changes/improvements required? As I see it is still open.

kivikakk:
I'm currently sick and unable to review the PR. I will when I can.

blood-rogue:
any updates on this?

kivikakk:
No updates. With my limited energy I've been prioritising bug fixes and upstream compliance.

PR #305, opened 2023-05-17
Make `ComrakExtensionOptions` non-exhaustive

### LINKED ISSUE #303: `shortcodes` feature isn't additive in `ComrakExtensionOptions`
Enabling the `shortcodes` feature can break existing code that compiles without it. E.g.

```rust
fn main() {
    let _ = ComrakExtensionOptions {
        strikethrough: false,
        tagfilter: false,
        table: false,
        autolink: false,
        tasklist: false,
        superscript: false,
        header_ids: None,
        footnotes: false,
        description_lists: false,
        front_matter_delimiter: None,
    };
}
```

compiles fine. However, enabling the `shortcodes` feature would then cause it to hit a compiler error about missing the `shortcodes` field :crying_cat_face: 

Simply switching the `ComrakExtensionOptions` struct to be `#[non_exhaustive]` would be enough to fix the issue, and since `ComrakExtensionOptions` impls `Default` it should still be easy for people to construct it however they please

I'd be happy to make a PR if that seems like a sensible solution

CosmicHorrorDev:
Resolves #303

_Note: This is a breaking change, so it will require a breaking semver bump on next release_

I would also be willing to add builders for constructing all parts of `ComrakOptions` which is a bit more ergonomic to construct things with. All intermediate states of constructing `ComrakOptions` are valid, so they can just be implemented with setters on the structs instead of requiring a separate type

### DIFF
--- a/examples/s-expr.rs
+++ b/examples/s-expr.rs
@@ -74,18 +74,18 @@ fn iter_nodes<'a, W: Write>(
 fn dump(source: &str) -> io::Result<()> {
     let arena = Arena::new();
 
+    let mut extension = ExtensionOptions::default();
+    extension.strikethrough = true;
+    extension.tagfilter = true;
+    extension.table = true;
+    extension.autolink = true;
+    extension.tasklist = true;
+    extension.superscript = true;
+    extension.footnotes = true;
+    extension.description_lists = true;
+
     let opts = Options {
-        extension: ExtensionOptions {
-            strikethrough: true,
-            tagfilter: true,
-            table: true,
-            autolink: true,
-            tasklist: true,
-            superscript: true,
-            footnotes: true,
-            description_lists: true,
-            ..ExtensionOptions::default()
-        },
+        extension,
         ..Options::default()
     };
 
--- a/fuzz/fuzz_targets/all_options.rs
+++ b/fuzz/fuzz_targets/all_options.rs
@@ -8,37 +8,36 @@ use comrak::{
 };
 
 fuzz_target!(|s: &str| {
+    let mut extension = ExtensionOptions::default();
+    extension.strikethrough = true;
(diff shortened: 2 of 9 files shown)


tgross35:
I am not a maintainer (no connection to the project) but would you maybe consider adding it to all the config structs?

```
ComrakExtensionOptions
ComrakOptions
ComrakParseOptions
ComrakPlugins
ComrakRenderOptions
ComrakRenderPlugins
```

If there is to be a one-time breaking change for exhaustiveness, may as well apply it to everything that might want to add options in the future

kivikakk:
> If there is to be a one-time breaking change for exhaustiveness, may as well apply it to everything that might want to add options in the future

I agree. I'll rebase this shortly and add those before I merge.

kivikakk:
I've decided to leave the top-level `Options` struct out of this change for now.

CosmicHorrorDev:
@kivikakk are you fine with me adding builders for the non-exhaustive options to make them less verbose to construct?

kivikakk:
I'm just rebasing @YJDoc2's #292 now! :)

CosmicHorrorDev:
Oh I somehow missed that 😅

Thanks for the great library btw!

tgross35:
> Thanks for the great library btw!

Agreed, we apprecaite it!

kivikakk:
Thanks both, that means a lot! The contributors list is getting longer — thanks for all your help. 😊

digitalmoksha:
🤔 does this change require an update to the README?

CosmicHorrorDev:
What part do you think would need a change? I don't see any code examples that use options or plugins aside from passing `Options::default()`

digitalmoksha:
Ah you’re right. I was thinking the calling interface was changing.

Issue #307, opened 2023-05-24
Parity: Footnote labels are not encoded


digitalmoksha:
When including characters in a footnote name, such as an emoji, those characters are not encoded.  `cmark-gfm` does encode these.

`cmark_gfm (cmark-gfm 0.29.0.gfm.10)`

```
second[^😄second]
[^😄second]: two
```

gives

```
<p>second<sup class="footnote-ref"><a href="#fn-%F0%9F%98%84second" id="fnref-%F0%9F%98%84second" data-footnote-ref>1</a></sup></p>
<section class="footnotes" data-footnotes>
<ol>
<li id="fn-%F0%9F%98%84second">
<p>two <a href="#fnref-%F0%9F%98%84second" class="footnote-backref" data-footnote-backref data-footnote-backref-idx="1" aria-label="Back to reference 1">↩</a></p>
</li>
</ol>
</section>
```

`comrak`

gives 

```
<p>second<sup class="footnote-ref"><a href="#fn-😄second" id="fnref-😄second" data-footnote-ref>1</a></sup></p>
<section class="footnotes" data-footnotes>
<ol>
<li id="fn-😄second">
<p>two <a href="#fnref-😄second" class="footnote-backref" data-footnote-backref data-footnote-backref-idx="1" aria-label="Back to reference 1">↩</a></p>
</li>
</ol>
</section>
```

digitalmoksha:
Looking at using the `percent_encoding` crate to encode utf8 characters

digitalmoksha:
Related, uppercase is stripped away in footnote names

```
third[^Test]
[^Test]: three
```

in `cmark-gfm` gives

```
<p>third<sup class="footnote-ref"><a href="#fn-Test" id="fnref-Test" data-footnote-ref>1</a></sup></p>
<section class="footnotes" data-footnotes>
<ol>
<li id="fn-Test">
<p>three <a href="#fnref-Test" class="footnote-backref" data-footnote-backref data-footnote-backref-idx="1" aria-label="Back to reference 1">↩</a></p>
</li>
</ol>
</section>
```

and `comrak` gives 

```
<p>third[^Test]</p>
```

We need a different normalization routine for footnote names

digitalmoksha:
Hmm, maybe it's `escape_href` in `html.rs` that I need...

digitalmoksha:
PR https://github.com/kivikakk/comrak/pull/308 addresses this

PR #308, opened 2023-05-25
Escape footnote name

digitalmoksha:
This escapes footnote names.  So `[^😄ref]` would yield the name `%F0%9F%98%84ref`

It also makes footnote names case-insensitive but case-preserving.  So if the footnote definition is `[^aB]: footnote`, you can reference it as `[^AB]` or `[^ab]`, etc.  The final name will be the definition name, `aB`.

Related to https://github.com/kivikakk/comrak/issues/307

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 target
 comrak-*
 .vscode
+.idea
--- a/src/html.rs
+++ b/src/html.rs
@@ -944,7 +944,9 @@ impl<'o> HtmlFormatter<'o> {
                     self.footnote_ix += 1;
                     self.output.write_all(b"<li")?;
                     self.render_sourcepos(node)?;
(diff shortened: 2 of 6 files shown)


digitalmoksha:
@kivikakk another step on the footnote journey...

There are two commits, one that fixes the encoding of footnote names, the other fixes the case preserving issue.

kivikakk (approved):

kivikakk:
Thanks — excellent. One thing you could consider is making a little purpose-built enum for the second argument, so that it looks more like `strings::normalize_label(&nfd.name, PreserveCase::DontPreserve)`, `strings::normalize_label(&nfd.name, PreserveCase::Preserve)` — or something like that, I'm not sure about what the best wording would be.

Issue #309, opened 2023-05-29
Backslash in a link


vpetrigo:
Hello!

Thank you for the awesome formatter. We experienced the following issue:
- backslash in a link is replaced after running the `comrak` tool

```
PS C:\Users\vladi> cat .\temp.md
[Link Text](#1\.-link)
PS C:\Users\vladi> comrak -t commonmark .\temp.md
[Link Text](#1.-link)
```

`[Link Text](#1\.-link)` is a valid link to a heading anchor within a wiki, but after formatting the navigation is broken.

**Version**:
```
comrak --version
comrak 0.18.0
```

kivikakk:
This is per spec; [see example](https://spec.commonmark.org/dingus/?text=%5BLink%20Text%5D(%231%5C.-link)).

vpetrigo:
Thank you for the example.
That did not explains why backslash is removed though if `[Link Text](#1\.-link)` and `[Link Text](#1.-link)` produces exact same Commonmark output.

Also the following backslash sequence in a link should produce [two different links](https://spec.commonmark.org/dingus/?text=%5BLink%20Text%5D(%231%5C.-link)%0A%5BLink%20Text%5D(%231%5C%5C%5C%5C.-link)):
```
PS C:\Users\vladi> cat .\temp.md
[Link Text](#1\.-link)
[Link Text](#1\\\\.-link)
PS C:\Users\vladi> comrak -t commonmark .\temp.md
[Link Text](#1.-link)
[Link Text](#1.-link)
```

while `comrak` reformats them to be the same.

vpetrigo:
> see examples 22 and 23, where the backslash is not included in the output HTML

Sorry, my bad. Quite unfamiliar with Commonmark specs yet. 😅

I'll check what can be done to produce proper output for links with backslashes.

Nulllix:
Maybe you should add special comments where you can wrap markdown text that comrak shouldn't touch. As is done in [clang-format](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#disabling-formatting-on-a-piece-of-code) (see Disabling Formatting on a Piece of Code) and other code formatting tools.

For example:
```
<!--- comrak off -->
[Link Text](#1\.-link)
[Link Text](#1\\\\.-link)
<!--- comrak on -->
```

Comrak will find comments like `comrak off/on`, remove them, and leave the text inside unformatted.

kivikakk:
[...]

PR #317, opened 2023-06-08
Fix backslash in a link issue

vpetrigo:
Resolves: #309

- Added skipping of a backslash in an escape sequence


kivikakk:
Thank you — this looks good to me.

PR #318, opened 2023-06-14
Sync with cmark-gfm-0.29.0.gfm.7

digitalmoksha:
Most of the changes from `gfm.6` to `gfm.7` have been previously incorporated.  This adds one minor thing that was missing, and fixes footnote handling to properly pass the `regressions.txt` tests, which should be passing at this version.

### DIFF
--- a/src/parser/inlines.rs
+++ b/src/parser/inlines.rs
@@ -1244,6 +1244,8 @@ impl<'a, 'r, 'o, 'd, 'i, 'c, 'subj> Subject<'a, 'r, 'o, 'd, 'i, 'c, 'subj> {
             let mut text = String::new();
             let mut sibling_iterator = bracket_inl_text.following_siblings();
 
+            self.pos = initial_pos;
+
             // Skip the initial node, which holds the `[`
             sibling_iterator.next().unwrap();
 
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -1759,6 +1759,13 @@ impl<'a, 'o, 'c> Parser<'a, 'o, 'c> {
         let mut ix = 0;
         Self::find_footnote_references(self.root, &mut map, &mut ix);
 
+        if !map.is_empty() {
+            // In order for references to be found inside footnote definitions,
+            // such as `[^1]: another reference[^2]`,
+            // the node needed to remain in the AST. Now we can remove them.
+            Self::cleanup_footnote_definitions(self.root);
+        }
+
         if ix > 0 {
             let mut v = map.into_values().collect::<Vec<_>>();
             v.sort_unstable_by(|a, b| a.ix.cmp(&b.ix));
@@ -1783,7 +1790,6 @@ impl<'a, 'o, 'c> Parser<'a, 'o, 'c> {
     ) {
         match node.data.borrow().value {
             NodeValue::FootnoteDefinition(ref nfd) => {
-                node.detach();
                 map.insert(
                     strings::normalize_label(&nfd.name, Case::DontPreserve),
                     FootnoteDefinition {
@@ -1840,6 +1846,19 @@ impl<'a, 'o, 'c> Parser<'a, 'o, 'c> {
             label.insert_str(0, "[^");
             label.push(']');
             ast.value = NodeValue::Text(label);
+        }
+    }
+
+    fn cleanup_footnote_definitions(node: &'a AstNode<'a>) {
+        match node.data.borrow().value {
+            NodeValue::FootnoteDefinition(_) => {
+                node.detach();
+            }
+            _ => {
+                for n in node.children() {
+                    Self::cleanup_footnote_definitions(n);
+                }
+            }
         }
     }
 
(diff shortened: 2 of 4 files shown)


digitalmoksha:
@kivikakk you're right, you already added just about everything from .6 to .7 💯 .  I found one minor missing change (the re-addition of `source` in `blocktagname`).

And I fixed some failures that should be working for this version.

Add anything that I may have missed for this version upgrade.  But we might be able to call this particular upgrade done? 🤔 🤞

kivikakk (approved):

kivikakk:
Thanks so much for this as always, @digitalmoksha!

> But we might be able to call this particular upgrade done? 🤔 🤞

Having had a look through the diffs, I think so! I'm now looking forward from here:

* 7–8 is adding the quadratic fuzzer (which we already have) and internal changes (which we can ignore).
* 8—9 adds sourcepos for autolinked URLs (we currently lack), footnote rendering changes (we have!), more internal changes.
* 9–10 does some counting and bookkeeping which fix a quadratic perf issue or two, and also makes it so directly nested emphasis doesn't render. This is the currently outstanding spec failure.
* 10—11 appears to back out a fair bit of the counting from 9—10 and does it a different way, so we should be careful not to spend time and effort on 9—10 and instead consider https://github.com/github/cmark-gfm/compare/0.29.0.gfm.9...0.29.0.gfm.11 in one go. There's some additions to the quadratic fuzzer we should incorporate too.

Everything remaining is pretty minor, and I'm happy to divide it up in whatever way you like.

kivikakk:
> Would you be up for tackling 8-9, since you recently added the sourcepos code and know auto link.

I'd be happy to! Sometime in the next day or two.


> I have to admit I'm a little disappointed in the nested emphasis code - that directly breaks running against the CommonMark specs. But I haven't yet looked at _why_ they did that.

It's a bit of an interesting one. I don't really understand the motivation myself: https://github.com/github/cmark-gfm/commit/5c75d23a02c52ee8b226ce53e3fb5339e4593757

There are some changes to how the tight/non-tight distinction is respected in CommonMark/plaintext output, which I would assume is actually responsible for the speed up mentioned in the commit message. The HTML output timing didn't change because there was no $\mathcal{O}(n^2)$ behaviour there to begin with.

We had to deal with this change in the `cmark-gfm` branch of Commonmarker, too, and the rationale is still not clear: https://github.com/gjtorikian/commonmarker/pull/236

kivikakk:
In this respect, it's kind of funny — `cmark-gfm` doesn't actually pass CommonMark any more, nor does it pass the [GitHub Flavored Markdown Spec as last published](https://github.github.com/gfm/), since the last version of that was published by me 🙃

The `cmark-gfm` test suite passes, but only because they've modified the spec file used by the suite in the repo: https://github.com/github/cmark-gfm/commits/master/test/spec.txt This file _should_ be identical to the one used to produce the HTML spec above (given it's the GFM spec reference!), but isn't.

digitalmoksha:
> It's a bit of an interesting one. I don't really understand the motivation myself: https://github.com/github/cmark-gfm/commit/5c75d23a02c52ee8b226ce53e3fb5339e4593757

Yeah it's bizarre, there doesn't seem to be a real reason to change the HTML renderer, and it breaks compatibility.  Not cool. Same here, had to deal with it when upgrading `commonmarker`, https://gitlab.com/gitlab-org/gitlab/-/merge_requests/121602.

> In this respect, it's kind of funny — cmark-gfm doesn't actually pass CommonMark any more, nor does it pass the [GitHub Flavored Markdown Spec as last published](https://github.github.com/gfm/), since the last version of that was published by me

Yeah, funny/not funny 😄.  I'm against deviating from the CommonMark spec.  It's bad enough we're not using 0.30 (I know, there aren't _that_ many real changes, but it's been 2 years).

digitalmoksha:
> These issues in cmark-gfm look fairly fixable (looks like some rewinds not being catered for correctly), and it doesn't seem to have any issues with smart punctuation. Our fix will be a bit hairier. I'll think on it.

Oof - I see how that can get complicated real fast.

I've been able to put together the `9->11` PR (which I'll put up here in a bit).  All the tests pass locally, including those in `script/cibuild`.

So we might be able to go ahead and say that we're synced up to 11, while the autolink is worked on.  As you said, we've already got the footnote changes, so I don't really see anything else important in `8->9`.

Interesting, I don't actually see any tests in `cmark-gfm` for `sourcepos`

kivikakk:
> I've been able to put together the `9->11` PR (which I'll put up here in a bit). All the tests pass locally, including those in `script/cibuild`.
> 
> So we might be able to go ahead and say that we're synced up to 11, while the autolink is worked on. As you said, we've already got the footnote changes, so I don't really see anything else important in `8->9`.

Amazing, thank you so much! And yes, I think that's probably the way to go.

> Interesting, I don't actually see any tests in `cmark-gfm` for `sourcepos`

There are some; they're kind of hidden since they're part of the API tests. See https://github.com/github/cmark-gfm/blob/2d65cd3c4bfbbdddc7accefc76392c16bb0cfb6d/api_test/main.c#L996-L1130.

PR #319, opened 2023-06-14
Sync with cmark-gfm-0.29.0.gfm.11

digitalmoksha:
This syncs with [cmark-gfm-0.29.0.gfm.11](https://github.com/github/cmark-gfm/releases/tag/0.29.0.gfm.11) release

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Collective](https://opencollective.com/comrak/all/badge.svg?label=financial+cont
 [![crates.io version](https://img.shields.io/crates/v/comrak.svg)](https://crates.io/crates/comrak)
 [![docs.rs](https://docs.rs/comrak/badge.svg)](https://docs.rs/comrak)
 
-Rust port of [github's `cmark-gfm`](https://github.com/github/cmark). _Currently synced with release `0.29.0.gfm.11`_
+Rust port of [github's `cmark-gfm`](https://github.com/github/cmark). *Currently synced with release `0.29.0.gfm.11`*.
 
 - [Installation](#installation)
 - [Usage](#usage)
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Collective](https://opencollective.com/comrak/all/badge.svg?label=financial+cont
 [![crates.io version](https://img.shields.io/crates/v/comrak.svg)](https://crates.io/crates/comrak)
 [![docs.rs](https://docs.rs/comrak/badge.svg)](https://docs.rs/comrak)
 
-Rust port of [github's `cmark-gfm`](https://github.com/github/cmark).
+Rust port of [github's `cmark-gfm`](https://github.com/github/cmark). _Currently synced with release `0.29.0.gfm.11`_
 
 - [Installation](#installation)
 - [Usage](#usage)
--- a/fuzz/fuzz_targets/quadratic.rs
+++ b/fuzz/fuzz_targets/quadratic.rs
@@ -2,7 +2,8 @@
 #![feature(int_roundings)]
 #![no_main]
 use comrak::{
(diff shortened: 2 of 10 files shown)


digitalmoksha:
@kivikakk I haven't updated `fuzz/fuzz_quadratic.c` or created `fuzz/fuzz_quadratic_brackets.c` yet.

I also have a small doubt on `render.c` in https://github.com/github/cmark-gfm/compare/0.29.0.gfm.10...0.29.0.gfm.11.  I don't think we need it, or at least I can't find where we would incorporate the changes 🤔

digitalmoksha:
Woohoo, green pipeline!

digitalmoksha:
So I think I've made the necessary changes to `fuzz/quadratic.rs`.

Looking at how you're using `Markdown::Markdown`, `Markdown::Sandwich`, and `Markdown::Tree`, I'm guessing we would want a `Markdown::Brackets`, rather than creating an entirely separate `quadratic_brackets.rs` file.

Though I admit my eyes are glazing over right now, and I'm not sure how it should be coded. I'm quite fuzzy on the fuzzer 😬

kivikakk (approved):

kivikakk:
Amazing! Thank you so much!

> Looking at how you're using `Markdown::Markdown`, `Markdown::Sandwich`, and `Markdown::Tree`, I'm guessing we would want a `Markdown::Brackets`, rather than creating an entirely separate `quadratic_brackets.rs` file.

I'm not sure myself—I didn't write it either. It looks to me like `Markdown::Sandwich` might actually be that? [`fuzz_quadratic_brackets.c`](https://github.com/github/cmark-gfm/blob/master/fuzz/fuzz_quadratic_brackets.c) seems to do something very similar—there's a section repeated some number of times before and after.

> I also have a small doubt on `render.c` in [github/cmark-gfm@0.29.0.gfm.10...0.29.0.gfm.11](https://github.com/github/cmark-gfm/compare/0.29.0.gfm.10...0.29.0.gfm.11). I don't think we need it, or at least I can't find where we would incorporate the changes 🤔

This is actually making it clear that there's another difference between Comrak and `cmark-gfm`, perhaps since that change. See:

```console
$ printf '3. x\n5. z' | cargo run -- --to commonmark
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/comrak --to commonmark`
3.  x
5.  z

$ printf '3. x\n5. z' | build/src/cmark-gfm -t commonmark
3.  x
4.  z
```

This difference isn't really able to be captured in the spec, since it only checks HTML output. You _could_ consider Comrak's roundtripping here more information-preserving, and it's not clear that our approach is worse.

You've done really well. Thank you. I'm happy to merge this now if you are.

kivikakk:
(Also, don't worry about the conflict, I'll resolve it when I merge.)

digitalmoksha:
> You could consider Comrak's roundtripping here more information-preserving, and it's not clear that our approach is worse.

Hmm, you're right.  But I think the main reason that I would use the `commonmark` output option would be to cleanup some text. I think in that case it would be preferable to have the items renumbered.

But it certainly could be looked at either way.  I went ahead and opened https://github.com/kivikakk/comrak/issues/323, because I found a problem with the numbering when using task lists.  Neither of these are a big deal since the resulting commonmark is valid.

> It looks to me like Markdown::Sandwich might actually be that?

I think you're probably right. 😌 

> I'm happy to merge this now if you are.

Let's do it! 🚀

kivikakk:
:tada: :tada: :tada: Amazing! Thank you so much! We're finally CI green again! 🍏

digitalmoksha:
Hey thanks for working through these with me so fast!

PR #320, opened 2023-06-16
Rename `ComrakFoo` types to just `Foo` for easier usage

tgross35:
There are a few types whose names start with `Comrak`:

```
ComrakExtensionOptions
ComrakOptions
ComrakParseOptions
ComrakPlugins
ComrakRenderOptions
ComrakRenderPlugins
```

This sort of naming is somewhat redundant - they are longer & clunky to work with, even though they are unlikely to conflict with anything. This PR removes the `Comrak` prefix so they are more ergonomic. Any users who do have conflicts can use e.g. `comrak::Plugins`.

This PR also adds aliases for the old type names, so no workflows are affected.

### DIFF
--- a/examples/update-readme.rs
+++ b/examples/update-readme.rs
@@ -1,10 +1,10 @@
-// Update the "comrak --help" text in Comrak's own README.
+// Update the "comrak --help" text in 's own README.
 
 use std::fmt::Write;
 use std::str;
 
 use comrak::nodes::{AstNode, NodeValue};
-use comrak::{format_commonmark, parse_document, Arena, ComrakOptions};
+use comrak::{format_commonmark, parse_document, Arena, Options};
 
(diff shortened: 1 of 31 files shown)


kivikakk:
Thanks, this is a very tidy change. I'll merge it when I'm next on the project!

examples/update-readme.rs
```diff
@@ -1,10 +1,10 @@
-// Update the "comrak --help" text in Comrak's own README.
+// Update the "comrak --help" text in 's own README.
```
kivikakk, line 1: A reminder to me to fix this up pre-merge and do another once-over.

```suggestion
// Update the "comrak --help" text in Comrak's own README.
```

tgross35:
Thanks!

PR #322, opened 2023-06-17
add PartialEq and Eq derive for Ast and its components

### LINKED ISSUE #211: Implementing PartialEq for Ast
I'm writing a program that needs to translate the comrak AST into a different AST, that a WYSIWYG editor (prosemirror) understands, and back again. I was about to write some tests for this translation, but I found that the [`Ast`](https://docs.rs/comrak/latest/comrak/nodes/struct.Ast.html) struct doesn't implement PartialEq:

```
error[E0369]: binary operation `==` cannot be applied to type `Ast`
   --> src/prosemirror.rs:105:9
    |
105 |         assert_eq!(*expected_comrak.data.borrow(), *actual_comrak.data.borrow());
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |         |
    |         Ast
    |         Ast
    |
    = note: this error originates in the macro `assert_eq` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Maybe there is there a reason for this that I'm missing? If not, would you consider a pull request that adds this? Would be happy to contribute the change.

I did notice there is a [`same_node` function on `Node`](https://docs.rs/comrak/latest/comrak/arena_tree/struct.Node.html#method.same_node) but it seems this is for checking if two nodes point to literally the same data.

PS. Thanks for maintaining this library!

YJDoc2:
Resolves #211

This adds PartialEq and Eq derive for Ast. As you had said you'd be happy to accept a PR on this, opening this. In case you have some other direction for this, feel free to update/close.

### DIFF
--- a/src/nodes.rs
+++ b/src/nodes.rs
@@ -8,7 +8,7 @@ use std::convert::TryFrom;
 use crate::parser::shortcodes::NodeShortCode;
 
 /// The core AST node enum.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum NodeValue {
     /// The root of every CommonMark document.  Contains **blocks**.
     Document,
@@ -154,7 +154,7 @@ pub enum NodeValue {
 }
 
 /// Alignment of a single table cell.
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub enum TableAlignment {
     /// Cell content is unaligned.
     None,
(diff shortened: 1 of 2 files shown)


YJDoc2:
Hey @kivikakk , I haven't made any code changes which can make the tests fail, and they seem to fail on current master as well. Can you help a little bit figuring out the issue?

kivikakk:
Hi again @YJDoc2, thanks so much for this!

Not to worry, this is an expected failure. @digitalmoksha's latest PR at #319 gets CI green again. Your changes are all good 👍

Issue #296, opened 2023-04-06
Parity: Autolink works incorrectly with links inside wiki style link


digitalmoksha:
Given the example

```
- [[linked-resource]]
- [[link-text|linked-resource]]
- [[http://example.com]]
- [[link-text|http://example.com/pdfs/gollum.pdf]]
- [[images/example.jpg]]
- [[http://example.com/images/example.jpg]]
```

comrak generates

```
<ul>
<li>[[linked-resource]]</li>
<li>[[link-text|linked-resource]]</li>
<li>[[<a href="http://example.com%5D%5D">http://example.com]]</a></li>
<li>[[link-text|<a href="http://example.com/pdfs/gollum.pdf%5D%5D">http://example.com/pdfs/gollum.pdf]]</a></li>
<li>[[images/example.jpg]]</li>
<li>[[<a href="http://example.com/images/example.jpg%5D%5D">http://example.com/images/example.jpg]]</a></li>
</ul>
```

Notice how the auto linking is incorrectly recognizing the links.

digitalmoksha:
For comparison with other parsers.  All except `cmark-gfm` properly recognize the boundary of the link, leaving the `[[` and `]]` intact. `cmark-gfm` don't recognize them as links at all.

**cmark-gfm**

```
<ul>
<li>[[linked-resource]]</li>
<li>[[link-text|linked-resource]]</li>
<li>[[http://example.com]]</li>
<li>[[link-text|http://example.com/pdfs/gollum.pdf]]</li>
<li>[[images/example.jpg]]</li>
<li>[[http://example.com/images/example.jpg]]</li>
</ul>
```

**markdown-rs**

```
<ul>
<li>[[linked-resource]]</li>
<li>[[link-text|linked-resource]]</li>
<li>[[<a href="http://example.com">http://example.com</a>]]</li>
<li>[[link-text|<a href="http://example.com/pdfs/gollum.pdf">http://example.com/pdfs/gollum.pdf</a>]]</li>
<li>[[images/example.jpg]]</li>
<li>[[<a href="http://example.com/images/example.jpg">http://example.com/images/example.jpg</a>]]</li>
</ul>
```

**markdown-it.rs**

```
<ul>
<li>[[linked-resource]]</li>
<li>[[link-text|linked-resource]]</li>
<li>[[<a href="http://example.com">example.com</a>]]</li>
[...]

PR #325, opened 2023-06-20
Fix autolink detection inside wiki style link brackets

digitalmoksha:
Related to https://github.com/kivikakk/comrak/issues/296

`cmark-gfm` does _not_ detect an autolink in `[[http://example.com]]`.  However by default `comrak` was.

This PR implements the `cmark-gfm` behavior.  It also introduces the option `relaxed-autolinks` to allow detection inside brackets, as many other markdown packages do.

### DIFF
--- a/src/tests/api.rs
+++ b/src/tests/api.rs
@@ -52,7 +52,8 @@ fn exercise_full_api() {
         parse: ParseOptions {
             smart: false,
             default_info_string: Some("abc".to_string()),
-            relaxed_tasklist_matching: true,
+            relaxed_tasklist_matching: false,
+            relaxed_autolinks: false,
         },
         render: RenderOptions {
             hardbreaks: false,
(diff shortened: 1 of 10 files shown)


digitalmoksha:
@kivikakk I think this takes care of the issue. Have a look when you get some time.

src/tests/api.rs
```diff
@@ -52,7 +52,8 @@ fn exercise_full_api() {
         parse: ParseOptions {
             smart: false,
             default_info_string: Some("abc".to_string()),
-            relaxed_tasklist_matching: true,
+            relaxed_tasklist_matching: false,
```
digitalmoksha, line 55: I updated this because the other options/extensions are turned off, including the task list extension.  So it seems like that's the correct state to start off with.

kivikakk: Any value is fine; we don't check the output! The purpose of this file was to notice when the API changed shape in any way, so I'd notice before I made a semver-breaking release.

I'll rewrite these to use the builders in a moment; there's no point to this test as written, since non-exhaustive means the structs can't be made this way from outside the crate.

kivikakk (approved):
Thank you!

Issue #324, opened 2023-06-19
[RFC] CI for running Benchmarks


YJDoc2:
Given that there are quite a few enhancements requested (https://github.com/kivikakk/comrak/issues/291 and linked in its desc), there will be code changes that potentially change the processing path or structures to implement them. As this is a parsing library and has speed as a concern, there should be a way to track change in processing time along with commits.

Currently I think there are benchmarks in makefile, so we should probably allow running them on github itself through CI. One way can be to run on each push, other would be to run through comment.

Would a PR on this welcome? Any suggestions on how to do benchmarks apart from the makefile and/or approach?

Thanks!

YJDoc2:
I think that makes sense! We can have an initial run when PR is opened, and then ad-hoc runs through comments. The reporting comment can be sticky, so there is not too much noise, and as you suggest it can also mention the way to re-run and update.

I see that the bench in makefile o/p gives mean, median and stdev ; so should we report these raw and diff from master or just the diff? Also  should we also report the values for cmark-gfm to compare?

gjtorikian:
> so should we report these raw and diff from master or just the diff?

I think raw + diff makes sense here. If it's helpful, [here's an example](https://github.com/gjtorikian/commonmarker/issues/107#issuecomment-553214816) of running the complete program (though it sounds like you know where to look already :)).

 > Also should we also report the values for cmark-gfm to compare?
 
 I think cmark-gfm plus the two libraries mentioned in the README (`[pulldown-cmark](https://github.com/google/pulldown-cmark)` and [markdown-rs](https://github.com/wooorm/markdown-rs)) would be nice.

kivikakk:
I think such a change is a really good idea! Very much welcome.

My main concern is about measurement noise from the execution environment? The benchmark script uses the user-space time exclusively (and not real/wall-clock), which should exclude variance due to vCPU scheduling etc. in VMs like those used by Actions, so it's likely fine.

> The reporting comment can be sticky, so there is not too much noise, and as you suggest it can also mention the way to re-run and update.

This is very neat!

YJDoc2:
Great! I'll try to get a PR up this weekend.

Also what are your thoughts on owning the benchmarking script, instead of using that of cmark-gfm? We will still be using same approach of using the pro-git md files, but instead of just `time` we can use something else (maybe hyperfine?) to run the benchmarks, so we can maybe have better measurements and have o/p format as we need.

[...]

PR #326, opened 2023-06-24
Add CI for running benchmarks

YJDoc2:
Solves #324 

As discussed, I have added CI for running benchmarks each time a PR is opened, and subsequently, when anyone comments `/run-bench` on the PR. This will run the benchmarks and post the result table as a comment on the PR. On subsequent runs, the same comment will be edited and updated with new results.

~Because github needs the CI file to be present in default branch to run a CI, the bench CI won't run for this PR.~

The CI is failing here, because the comment post action needs write permission on PRs, which is not present for default github token provided to CI in this repo (?). I have opened a dummy PR [here](https://github.com/YJDoc2/comrak/pull/6) on my fork. you can see the posted results there, as well as try out the comment based invocation of the benchmarks.

I have also changed the benchmarking method as discussed. We used to simply clone the cmark-gfm and run the bench from there ; now I have copied the benchmark running script from there, and use hyperfine to run them. Now we also run benchmark on pulldown-cmark and cmark-gfm . Markdown-rs was also considered, but it is a library-only repo, so we will probably need to setup cargo project with simple binary which reads from stdin and runs the markdow-rs parse on it for running benchmarks on it. That is not done in this PR.

Some issues I have noticed :
- Benchmark CI takes ~5 minutes total to build and run benchmark on all repos. Major time is taken in building all of the binaries, the benchmarks only take ~1 minute total to run.
- The time measurement varies for each CI run considerably. Not sure if this is GH runner issue or what, because when running this locally, the measurements were consistent across several runs. What this would mean is that we cannot compare measurement from a previous CI run to current, but only the values in current CI run. This also means that the `relative` section in the measurement is most sensible for understanding difference and changes in speed. You can see the comment history of the measurement comment on my PR (linked above) to see what I mean by all this

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,0 +1,36 @@
+name: benchmarks
+
+on: 
+  pull_request:
+    types:
+      - opened
+      - reopened
+  issue_comment:
+    types:
+      - created
+
+jobs:
+  run_benchmarks:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    # run  either when pull request is opened or when comment body (only on pr) is /run-bench
+    if: (github.event_name == 'pull_request') || ((github.event.issue.pull_request != null) && github.event.comment.body == '/run-bench')
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: true
+      - name: Setup Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - name: Install hyperfine
+        run: cargo install hyperfine
+      - name: Install cmake
+        run: sudo apt-get update && sudo apt-get install cmake -y
+      - name: Build Binaries
+        run: make binaries
+      - name: Run Benchmarks
+        run: make bench-all
+      - name: Post result comment
+        uses: mshick/add-pr-comment@v2
+        with:
+          message-path: bench-output.md
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,7 @@
+ROOT:=$(shell git rev-parse --show-toplevel)
+COMMIT:=$(shell git rev-parse --short HEAD)
+MIN_RUNS:=25
+
 src/scanners.rs: src/scanners.re
 	re2rust -W -Werror -i --no-generation-date -o $@ $<
 	cargo fmt
@@ -5,3 +9,35 @@ src/scanners.rs: src/scanners.re
 bench:
 	cargo build --release
 	(cd vendor/cmark-gfm/; make bench PROG=../../target/release/comrak)
+
+binaries: build-comrak-branch build-comrak-master build-cmark-gfm build-pulldown-cmark
+
+build-comrak-branch:
+	cargo build --release
+	cp ${ROOT}/target/release/comrak ${ROOT}/benches/comrak-${COMMIT}
+
+build-comrak-master:
+	git clone https://github.com/kivikakk/comrak.git --depth 1 --single-branch ${ROOT}/vendor/comrak || true
+	cd ${ROOT}/vendor/comrak && \
+	cargo build --release && \
+	cp ./target/release/comrak ${ROOT}/benches/comrak-main
+
+build-cmark-gfm:
+	cd ${ROOT}/vendor/cmark-gfm && \
+	make && \
+	cp build/src/cmark-gfm ${ROOT}/benches/cmark-gfm
+
+build-pulldown-cmark:
+	cd ${ROOT}/vendor/pulldown-cmark && \
+	cargo build --release && \
+	cp target/release/pulldown-cmark ${ROOT}/benches/pulldown-cmark
+
+bench-comrak: build-comrak-branch
+	git clone https://github.com/progit/progit.git ${ROOT}/vendor/progit || true > /dev/null
+	cd benches && \
+	hyperfine --prepare 'sudo sync; echo 3 | sudo tee /proc/sys/vm/drop_caches' --warmup 3 --min-runs ${MIN_RUNS}  -L binary comrak-${COMMIT} './bench.sh ./{binary}'
+
+bench-all: binaries
+	git clone https://github.com/progit/progit.git ${ROOT}/vendor/progit || true > /dev/null
+	cd benches && \
+	hyperfine --prepare 'sudo sync; echo 3 | sudo tee /proc/sys/vm/drop_caches' --warmup 3 --min-runs ${MIN_RUNS}  -L binary comrak-${COMMIT},comrak-main,pulldown-cmark,cmark-gfm './bench.sh ./{binary}' --export-markdown ${ROOT}/bench-output.md
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ bench:
 	cargo build --release
 	(cd vendor/cmark-gfm/; make bench PROG=../../target/release/comrak)
 
-binaries: build-comrak-branch build-comrak-master build-cmark-gfm build-pulldown-cmark
+binaries: build-comrak-branch build-comrak-master build-cmark-gfm build-pulldown-cmark build-markdown-it
 
 build-comrak-branch:
 	cargo build --release
@@ -27,6 +27,11 @@ build-cmark-gfm:
 	make && \
 	cp build/src/cmark-gfm ${ROOT}/benches/cmark-gfm
 
+build-markdown-it:
+	cd ${ROOT}/vendor/markdown-it && \
+	cargo build --release && \
+	cp target/release/markdown-it ${ROOT}/benches/markdown-it
+
 build-pulldown-cmark:
 	cd ${ROOT}/vendor/pulldown-cmark && \
 	cargo build --release && \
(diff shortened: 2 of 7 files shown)


benches/bench.sh
```diff
@@ -0,0 +1,8 @@
+#! /bin/bash
+
+PROG=$1
+ROOTDIR=$(git rev-parse --show-toplevel)
+
+for lang in ar az be ca cs de en eo es es-ni fa fi fr hi hu id it ja ko mk nl no-nb pl pt-br ro ru sr th tr uk vi zh zh-tw; do \
+    cat $ROOTDIR/vendor/progit/$lang/*/*.markdown | $PROG > /dev/null
```
YJDoc2, line 7: We don't use `seq 1..10` loop here like in original file from cmkar-gfm because instead the script is run several times by hyperfine itself

Makefile
```diff
@@ -1,7 +1,43 @@
+ROOT:=$(shell git rev-parse --show-toplevel)
+COMMIT:=$(shell git rev-parse --short HEAD)
+MIN_RUNS:=25
```
YJDoc2, line 3: To decide this value, I started with 10 runs,  and then checked 50 runs. the difference in measurements were not that much, even the std dev hadn't changed that much from 10 runs to 50. So I settled on 25.

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,36 @@
...
+    types:
+      - created
+
+jobs:
+  run_benchmarks:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    # run  either when pull request is opened or when comment body (only on pr) is /run-bench
+    if: (github.event_name == 'pull_request') || ((github.event.issue.pull_request != null) && github.event.comment.body == '/run-bench')
```
YJDoc2, line 18: I can change the comment body from `/run-bench` to something else if you want.

YJDoc2:
@kivikakk @digitalmoksha Please take a look when possible :)

Makefile
```diff
@@ -1,7 +1,43 @@
...
+	cd ${ROOT}/vendor/cmark-gfm && \
+	make && \
+	cp build/src/cmark-gfm ${ROOT}/benches/cmark-gfm
+
+build-pulldown-cmark:
+	cd ${ROOT}/vendor/pulldown-cmark && \
+	cargo build --release && \
+	cp target/release/pulldown-cmark ${ROOT}/benches/pulldown-cmark
+
+bench-comrak: build-comrak-branch
```
digitalmoksha, line 35: wdyt about adding [markdown-it.rs](https://github.com/rlidwka/markdown-it.rs) to the mix? It would be nice to have [markdown-rs](https://github.com/wooorm/markdown-rs), but as you said, at the moment there is not a command line version.

One thing we have to remember is that this really isn't testing exactly apples to apples. For example `markdown-rs` doesn't support footnotes and task lists directly yet. Though some [plugins](https://github.com/chrisjsewell/markdown-it-plugins.rs) have just been added for that.

A follow-on PR might have a small harness program for the Rust implementations, which would link in all the different `lib` versions.  The harness would take a switch, and then run the specified implementation, with the proper switches set to make them all use similar functionality.

digitalmoksha (commented):
@YJDoc2 this looks pretty good from my angle! Just had a few comments.

kivikakk:
This is looking amazing! Thank you so much.

I echo @digitalmoksha's comments about the use of `sudo` — is there usually a measurable difference between clearing the caches for our purposes? I imagine the warmup runs will remove a lot of the startup variation even without doing so.

> * The time measurement varies for each CI run considerably. Not sure if this is GH runner issue or what, because when running this locally, the measurements were consistent across several runs. What this would mean is that we cannot compare measurement from a previous CI run to current, but only the values in current CI run. This also means that the `relative` section in the measurement is most sensible for understanding difference and changes in speed. You can see the comment history of the measurement comment on my PR (linked above) to see what I mean by all this

Yes, that makes a lot of sense. I'm hopeful the relative data itself is accurate enough, often enough — the inter-run differences are probably caused by being scheduled onto hosts with vastly different loads at that time, but there's no predicting when that might happen during a run.

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Install cmake
+        run: sudo apt-get update && sudo apt-get install cmake -y
+      - name: Build Binaries
+        run: make binaries
+      - name: Run Benchmarks
+        run: make bench-all
+      - name: Post result comment
+        uses: mshick/add-pr-comment@v2
+        with:
+          message-path: bench-output.md
```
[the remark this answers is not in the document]
YJDoc2: Hey, good idea! I'll add a line in the comment showing when the CI was run date+time (GMT). Thanks for the suggestion :+1:

Makefile
YJDoc2: Sure, I didn't know about it ; so will take a look and try adding it as well.

YJDoc2:
Hey @kivikakk @digitalmoksha , thanks for the review :)

>Hmm, do we really need to use sudo? I tried make bench-comrak locally, and hyperfine just hung during warmup, and no idea why. Turns out sudo was asking for a password. I don't really feel comfortable giving a bench mark program sudo access.

>I echo @digitalmoksha's comments about the use of sudo — is there usually a measurable difference between clearing the caches for our purposes? I imagine the warmup runs will remove a lot of the startup variation even without doing so.

I added the cache clearing step because otherwise hyperfine gave a warning that "first run is an outlier, which might be an effect of cache". Also when I tried without cache clearing, the measurements were considerably different. (consistent, but different). That said, I don't like the sudo use either, especially for benchmarks, so I'll try to figure out a workaround. Maybe increasing the number of warmup runs might solves the issue. Will get back after updating this.

Thanks :)

YJDoc2:
@kivikakk @digitalmoksha I have removed sudo and added markdown-it rs in comparisons. The benchmark value change from using the cache drops, but it is consistent within itself, so it should be fine.  I have also added timestamp in the comment. Please take a look.

I'll see why the CI is failing...

README.md
```diff
@@ -197,6 +197,33 @@ assert_eq!(
...
+For running benchmarks, you will need to [install hyperfine](https://github.com/sharkdp/hyperfine#installation) and optionally cmake.
+
+If you want to just run benchmark for comrak, with current state of repo, you can simply run
+```bash
+make bench-comrak
+```
+
+This will build comrak in release mode, and run benchmark on it. You will see the time measurements as reported by hyperfine in the console.
+
+Makefile also provides a way to run benchmarks for comark current state (with your changes), comrak main branch, cmark-gfm and pulldown-cmark. For this you will need to install cmake. After that make sure that you have set-up the git submodules. In case you have not installed submodules when cloning, you can do it by running
```
digitalmoksha, line 211: Did you want to add a mention of `markdown-it.rs` here?

I added links to the repos as well, hope you don't mind.

```suggestion
Makefile also provides a way to run benchmarks for `comrak` current state (with your changes), `comrak` main branch, [`cmark-gfm`](https://github.com/github/cmark-gfm), [`pulldown-cmark`](https://github.com/raphlinus/pulldown-cmark) and [`markdown-it.rs`](https://github.com/rlidwka/markdown-it.rs). For this you will need to install `cmake`. After that make sure that you have set-up the git submodules. In case you have not installed submodules when cloning, you can do it by running
```

README.md
```diff
@@ -197,6 +197,33 @@ assert_eq!(
...
+```
+
+After this is done, you can run 
+```bash
+make bench-all
+```
+
+which will run benchmarks across all, and report the time take by each as well as relative time.
+
+Apart from this, CI is also setup for running benchmarks when a pull request is first opened. It will comment the results on the pull request in a tabular format comparing the 4 versions. After that you can manually trigger this CI by commenting `/run-bench` on the PR, this will update the existing comment with new results. Note Benchmarks won't be automatically run on each push.
```
digitalmoksha, line 223: Just a suggestion

```suggestion
Apart from this, CI is also setup for running benchmarks when a pull request is first opened. It will add a comment with the results on the pull request in a tabular format comparing the 5 versions. After that you can manually trigger this CI by commenting `/run-bench` on the PR, this will update the existing comment with new results. Note benchmarks won't be automatically run on each push.
```

README.md
```diff
@@ -197,6 +197,33 @@ assert_eq!(
      </ol>\n");
 ```
 
+## Benchmarking
+
+For running benchmarks, you will need to [install hyperfine](https://github.com/sharkdp/hyperfine#installation) and optionally cmake.
+
+If you want to just run benchmark for comrak, with current state of repo, you can simply run
```
digitalmoksha, line 204: Minor suggestion

```suggestion
If you want to just run the benchmark for `comrak`, with the current state of the repo, you can simply run
```

digitalmoksha:
@YJDoc2 this look good to me, nice job.

kivikakk:
I'll look into the CI failure this afternoon with a view to merging this.

YJDoc2:
Hey, sorry I wasn't able to update on this last week.

> I'll look into the CI failure this afternoon with a view to merging this.

Thanks a lot. Is it ok if I rebase it with main and force push? Maybe the CI is expecting some new things in master that aren't pulled in this branch.

YJDoc2:
@kivikakk I have updated the proc-macro2 version, as that was causing the build failure on nightly.
Related issue https://github.com/dtolnay/proc-macro2/issues/356 .
I haven't updated anything else in cargo lock.

kivikakk (approved):
Looks excellent. Tested all locally — thank you so much for your effort! 🤍

PR #328, opened 2023-06-28
Comma delimited info string language

### LINKED ISSUE #246: Option to treat `,` as a word delimiter for info-strings
Currently when parsing the info-string to determine the codeblock language the word up to the first space character is used:

https://github.com/kivikakk/comrak/blob/03238b81d0917acbea73a2d255603f8051e4dc04/src/html.rs#L490-L492

This causes issues with markdown such as in [the `regex` readme](https://github.com/rust-lang/regex/blob/master/README.md#usage-avoid-compiling-the-same-regex-in-a-loop) using the info-string `rust,ignore`, it is passed into the syntax highlighter as the string `rust,ignore` and applied as an attribute `class="language-rust,ignore"` on the element. Both rustdoc and github support the `,` character being a delimiter between the language and additional attributes (`rustdoc` actually supports more, but that's for back-compat, afaik only `[ ,]` is intended to be used).

CosmicHorrorDev:
Fixes #246 

Changes the syntect plugin to split off any text following a comma from the language tag to match what is supported by rustdoc and GitHub. This now accepts ```` ```rust,no_run```` as Rust code

### DIFF
--- a/src/tests/plugins.rs
+++ b/src/tests/plugins.rs
@@ -91,16 +91,22 @@ fn heading_adapter_plugin() {
 fn syntect_plugin() {
     let adapter = crate::plugins::syntect::SyntectAdapter::new("base16-ocean.dark");
 
-    let input = concat!("```rust\n", "fn main<'a>();\n", "```\n");
-    let expected = concat!(
-        "<pre style=\"background-color:#2b303b;\"><code class=\"language-rust\">",
-        "<span style=\"color:#b48ead;\">fn </span><span style=\"color:#8fa1b3;\">main</span><span style=\"color:#c0c5ce;\">",
-        "&lt;</span><span style=\"color:#b48ead;\">&#39;a</span><span style=\"color:#c0c5ce;\">&gt;();\n</span>",
-        "</code></pre>\n"
-    );
+    let cases = vec![
+        (
+            concat!("```rust\n", "fn main<'a>();\n", "```\n"),
+            concat!(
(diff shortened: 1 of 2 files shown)


src/tests/plugins.rs
```diff
@@ -91,16 +91,37 @@ fn heading_adapter_plugin() {
...
+        ),
+        (
+            // Language should still be highlighted when delimited by a comma
+            "\
+            ```rust,ignore\n\
+            fn main() {}\n\
+            ```\n\
+            ",
+            "\
+            <pre style=\"background-color:#2b303b;\"><code class=\"language-rust,ignore\">\
```
Nemo157, line 112: This still has `class="language-rust,ignore"`, so it wouldn't match anything looking for the class `language-rust`.

CosmicHorrorDev: I see now that this doesn't fit your use case. Changing that would differ from `cmark-gfm`'s output, but I should be able to add that functionality behind a parse option instead

I'll work on handling that today

Nemo157: It does match GitHub's behavior though, so maybe `cmark-gfm` has a bug 🤔

digitalmoksha: My guess is that GitHub uses the `full-info-string` option (which gives the full info string, not just the first word) and then parses the language out themselves. That's what we do.

CosmicHorrorDev: I'm not sure on how to proceed, so I'll just be awaiting more definitive feedback

kivikakk:
Merged with #752!

PR #331, opened 2023-07-17
new test wrap preserve link

jnnnnn:
This is a failing test for an issue I encountered: links with spaces in the link text get broken up and don't render correctly thereafter.

I've spent a few minutes but I can't see an obvious fix. I think parsing the link as a separate node and disabling wrap on the node is probably the way to go but I'd appreciate a pointer into the code for where to change this.

Thanks for looking and for a really fast Markdown formatter!


kivikakk:
Thanks for this very clear report, I appreciate it!

I'm a little unsure what the best thing to do is, here. Our behaviour (intentionally) matches upstream [cmark-gfm](https://github.com/github/cmark-gfm)'s exactly:

```console
$ echo 'This is [a link containing spaces that should not be wrapped](https://example.com) and then the text continues after.' | cmark-gfm -t commonmark --width 20
This is [a link
containing spaces
that should not be
wrapped](https://example.com)
and then the text
continues after.
$
```

Indeed, it's the same in the [reference CommonMark implementation, `cmark`](https://github.com/commonmark/cmark). While I do think this reduces its usefulness for round-tripping, it's not immediately clear if we should diverge from their behaviour, at least by default.

Changes would happen in [`comrak::cm`](https://github.com/kivikakk/comrak/blob/72b5209405e6177ff04577013daa7febd70fe615/src/cm.rs); searching for `render.width` shows you various places where the option is honoured. Your desired output would need us to disable wrapping so long as there's a link node somewhere in the current output ancestor chain.

An even nicer version of this would wrap at the _start_ of the link if the link was going to wrap otherwise. i.e.:

```diff
 12345678901234567890
-This is [a link
-containing
+This is
+[a link containing
 spaces](https://example.com).
```

jnnnnn:
Ah, it's just the VSCode markdown highlighter that doesn't highlight correctly.

image

PR #338, opened 2023-08-29
Use Nix for CI.

charlottia:


kivikakk:
Merging mostly so I can `nix develop` and work on Comrak again easily — remainder of any CI conversion (if any!) to be done later.

Issue #339, opened 2023-08-29
Benchmark PR job failing due to permissions


kivikakk:
See e.g. https://github.com/kivikakk/comrak/actions/runs/5958432069/job/16162659993?pr=337.

```
Run mshick/add-pr-comment@v2
  with:
    message-path: bench-output.md
    message-id: add-pr-comment
    refresh-message-position: false
    repo-owner: kivikakk
    repo-name: comrak
    repo-token: ***
    allow-repeats: false
    status: success
  env:
    CARGO_INCREMENTAL: 0
    CARGO_TERM_COLOR: always
Error: Resource not accessible by integration
```

This "Resource not accessible by integration" message has been coming up on a few projects lately where it didn't before; looks like Actions has changed their permissions model a bit.

digitalmoksha:
In particular it looks like a PR from a fork doesn't have the correct permissions.

https://github.com/mshick/add-pr-comment/issues/67 and https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

kivikakk:
Possible solution: https://github.com/mshick/add-pr-comment#proxy-for-fork-based-prs

kivikakk:
I've just made it non-failing for the PR as a whole for now.

PR #344, opened 2023-09-29
Use github/cmark-gfm submodule

digitalmoksha:
This shifts the `cmark-gfm` submodule back to `https://github.com/github/cmark-gfm.git`, due to the recent deletion of `https://github.com/kivikakk/cmark-gfm.git`

Tests currently pass this with version, which is at `0.29.0.13`.

We can still create another fork later if we need more control.


digitalmoksha:
@kivikakk @gjtorikian @charlottia what do you think about this? Or is there a better solution?

charlottia:
LGTM — thank you. I'll merge as-is; the benchmark CI failure is #339.

PR #347, opened 2023-11-12
Allow for Syntect to simply generate CSS classes

gjtorikian:
Syntect has the ability to [generate its syntax highlighting with CSS classes](https://github.com/trishume/syntect/blob/95df8850c9859c5403408cf8e7b5a2d7cc6cda2b/examples/synhtml-css-classes.rs#L96), rather than inline styles. 

I wanted to use this in Commonmarker, but I couldn't find an easy, non-breaking API to support this feature. I'm happy to rework or rearchitect it into anything that looks better. Technically this is still a non-1.0 library so breaking changes are permitted per sem ver, but, well, you know.... _people_.

### DIFF
--- a/examples/syntect.rs
+++ b/examples/syntect.rs
@@ -4,7 +4,7 @@ use comrak::plugins::syntect::SyntectAdapter;
 use comrak::{markdown_to_html_with_plugins, Options, Plugins};
 
 fn main() {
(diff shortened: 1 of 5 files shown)


kivikakk:
Thank you so much! :heart:

PR #349, opened 2023-11-29
Use text-align over deprecated align attribute for td and th

popekim:
align attribute is deprecated, and the text aligning of <td> and <th> elements should be done by text-align

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -916,13 +916,13 @@ impl<'o> HtmlFormatter<'o> {
 
                     match alignments[i] {
                         TableAlignment::Left => {
-                            self.output.write_all(b" align=\"left\"")?;
+                            self.output.write_all(b" style=\"text-align: left\"")?;
                         }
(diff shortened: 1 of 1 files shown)


gjtorikian:
I'll let @kivikakk have final say here. The goal of this library is:

> to model the upstream [cmark-gfm](https://github.com/github/cmark-gfm) as closely as possible in terms of code structure.

[cmark-gfm toggles this alignment with an option](https://github.com/github/cmark-gfm/blob/587a12bb54d95ac37241377e6ddc93ea0e45439b/extensions/table.c#L730-L739), so that may also be needed here. 

(Personally, damn the past and accept the future, LGTM)

kivikakk:
Yes, I'd be happy to accept an option to toggle this; I don't particularly want to break executable spec compat (see CI) or change output for all existing users.

kivikakk:
Closing since there's no option forthcoming! Thanks for your efforts here.

PR #359, opened 2024-01-23
Add a multiline blockquote extension

digitalmoksha:
Adds the ability to enclose multiple lines in a blockquote. Such as:

```
>>>
Paragraph one

Paragraph two
>>>
```

This is part of the GitLab Flavored Markdown (GLFM) specification, https://docs.gitlab.com/ee/user/markdown.html#multiline-blockquote

### DIFF
--- a/src/cm.rs
+++ b/src/cm.rs
@@ -377,6 +377,7 @@ impl<'a, 'o> CommonMarkFormatter<'a, 'o> {
             NodeValue::FootnoteReference(ref nfr) => {
                 self.format_footnote_reference(nfr.name.as_bytes(), entering)
             }
+            NodeValue::MultilineBlockQuote(..) => self.format_block_quote(entering),
         };
         true
     }
--- a/src/nodes.rs
+++ b/src/nodes.rs
@@ -480,7 +480,7 @@ impl NodeValue {
             NodeValue::FootnoteReference(..) => "footnote_reference",
             #[cfg(feature = "shortcodes")]
             NodeValue::ShortCode(_) => "shortcode",
-            NodeValue::MultilineBlockQuote(_) => "block_quote",
+            NodeValue::MultilineBlockQuote(_) => "multiline_block_quote",
         }
     }
 }
--- a/src/nodes.rs
+++ b/src/nodes.rs
@@ -7,6 +7,8 @@ use std::convert::TryFrom;
 #[cfg(feature = "shortcodes")]
 use crate::parser::shortcodes::NodeShortCode;
 
+use crate::parser::multiline_block_quote::NodeMultilineBlockQuote;
+
 /// The core AST node enum.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum NodeValue {
@@ -151,6 +153,19 @@ pub enum NodeValue {
     #[cfg(feature = "shortcodes")]
     /// **Inline**. An Emoji character generated from a shortcode. Enable with feature "shortcodes".
     ShortCode(NodeShortCode),
+
+    /// **Block**. A [multiline block quote](https://github.github.com/gfm/#block-quotes).  Spans multiple
+    /// lines and contains other **blocks**.
+    ///
+    /// ``` md
+    /// >>>
+    /// A paragraph.
+    ///
+    /// - item one
+    /// - item two
+    /// >>>
+    /// ```
+    MultilineBlockQuote(NodeMultilineBlockQuote),
 }
 
 /// Alignment of a single table cell.
@@ -391,6 +406,7 @@ impl NodeValue {
                 | NodeValue::TableRow(..)
                 | NodeValue::TableCell
                 | NodeValue::TaskItem(..)
+                | NodeValue::MultilineBlockQuote(_)
         )
     }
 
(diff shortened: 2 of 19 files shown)


digitalmoksha:
I still need to add an extension flag (right now it's fully hooked in).

And does it make sense make it a compile feature?

@kivikakk what do you think, would this be something that could be incorporated?

src/nodes.rs
```diff
@@ -464,6 +480,7 @@ impl NodeValue {
             NodeValue::FootnoteReference(..) => "footnote_reference",
             #[cfg(feature = "shortcodes")]
             NodeValue::ShortCode(_) => "shortcode",
+            NodeValue::MultilineBlockQuote(_) => "block_quote",
```
kivikakk, line 483: Happy to leave this as-is, but if you want to be able to roundtrip a blockquote's multilineness to XML[^xml], it'll need some distinction here.

[^xml]: which, let's face it, you probably don't rely on!

src/tests/api.rs
```diff
@@ -212,5 +212,6 @@ fn exercise_full_api() {
             let _: String = nfr.name;
             let _: u32 = nfr.ix;
         }
+        nodes::NodeValue::MultilineBlockQuote(_) => {}
```
kivikakk, line 215: Let's add detail of the fields (per above), so that we catch any API-breaking changes to the multiline block quote.

src/cm.rs
```diff
@@ -377,6 +377,7 @@ impl<'a, 'o> CommonMarkFormatter<'a, 'o> {
             NodeValue::FootnoteReference(ref nfr) => {
                 self.format_footnote_reference(nfr.name.as_bytes(), entering)
             }
+            NodeValue::MultilineBlockQuote(..) => self.format_block_quote(entering),
```
digitalmoksha, line 380: For now just outputting a regular block quote, as opposed to outputting the actual multiline format. I _think_ this makes sense, though I could see both ways.

src/parser/multiline_block_quote.rs
```diff
@@ -0,0 +1,9 @@
+/// The metadata of a multiline blockquote.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct NodeMultilineBlockQuote {
+    /// The length of the fence.
+    pub fence_length: usize,
+
+    /// ??? The indentation level of the code within the block.
```
kivikakk, line 7: Indentation level of the fence itself, in order to ensure that items within are correctly interpreted (e.g. per the "Marker can be indented up to 3 spaces" example).

src/cm.rs
kivikakk: Yes, this seems fine! If you want to be able to round-trip it, we can always add it back in later.

kivikakk (approved):
This looks awesome!

@digitalmoksha yes, I think we should definitely include this! I'd be inclined not to compile flag it, I think it's a handy option to have and there are increasingly Markdown dialects/input elements that support this syntax; no need to restrict it imo.

Let me know when you're done and I'll throw fuzzers at it for a while :)

src/nodes.rs
digitalmoksha: I originally had it as `multiline_block_quote`, and then second guessed myself.  Thanks, I'll change it back.

digitalmoksha:
> there are increasingly Markdown dialects/input elements that support this syntax

Really, I wasn't aware of any. Do you have an example?

kivikakk:
The one that comes to mind for me immediately is Discord: entering `>>>` at the start of a line will make the entire rest of the message fall under that blockquote. (There's no way to end it before the end of the message iirc. This behaviour only functions on mobile; on desktop the WYSIWYG editor takes over and produces the same effect.)

IIRC one or both of Slack and Mattermost also recognise this syntax, and I've seen people actually try it out elsewhere where it isn't supported!

src/parser/multiline_block_quote.rs
digitalmoksha: Spaced on updating that comment, thanks!

gjtorikian:
> IIRC one or both of Slack and Mattermost also recognise this syntax

FWIW, Slack for sure supports it.

digitalmoksha:
It looks Slack is the same as Discord, you can't turn off the block quote (at least I wasn't able to). You can't bracket it like a fenced code block.

kivikakk:
Yep, indeed! They don't share that aspect of it, but it's similar and intuitive enough (imo).

digitalmoksha:
I made the suggested changes and added the extension option `multline-block-quotes`

I think it's ready for final review.

kivikakk (approved):
LGTM! Thank you so much, @digitalmoksha! :heart: :shipit:

I'm going to run fuzzing on this for a few hours, and merge when done.

digitalmoksha:
Thanks for merging @kivikakk 🙇

digitalmoksha:
@kivikakk when you have time, would it be possible to cut a new release?

kivikakk:
Of course, will do so tonight!El 25 ene 2024, a la(s) 5:34, digitalMoksha ***@***.***> escribió:﻿
@kivikakk when you have time, would it be possible to cut a new release?

kivikakk:
[comrak 0.21.0](https://crates.io/crates/comrak/0.21.0) is now on crates.io! Thanks for your hard work, @digitalmoksha :heart:

digitalmoksha:
Thank you @kivikakk! I've updated my dingus, it's working like a charm. https://gitlab-org.gitlab.io/ruby/gems/gitlab-glfm-markdown/

Issue #361, opened 2024-01-25
Width options usage ?


Its-Just-Nans:
Hello,

How can I use the --width options in the CLI ?

With and without the options, the output isn't different ?!

```sh
cargo install comrak

comrak --gfm -o index_0.html README.md
comrak --gfm -o index_1.html --width 50 README.md

diff index_*.html
# no difference
```

kivikakk:
Presently the `--width` option only applies to the CommonMark output mode.

PR #362, opened 2024-01-26
Add support for width in html output

Its-Just-Nans:
Hello,

Here is a tiny PR to add `--width` support in HTML

- related to https://github.com/kivikakk/comrak/issues/361

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -31,6 +31,15 @@ pub fn format_document_with_plugins<'a>(
     output: &mut dyn Write,
     plugins: &Plugins,
 ) -> io::Result<()> {
+    if options.render.width > 0 {
+        output.write_all(b"<!DOCTYPE html>\n")?;
+        output.write_all(b"<html>\n<head> <meta charset=\"UTF-8\"><meta name=\"viewport\" content=\"width=device-width, initial-scale=1\"></head>")?;
+        let body_tag = format!(
+            "<body style=\"width: {}%; margin: auto\">",
+            options.render.width
+        );
+        output.write_all(body_tag.as_bytes())?;
+    }
     let mut writer = WriteWithLast {
         output,
         last_was_lf: Cell::new(true),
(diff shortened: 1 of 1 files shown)


kivikakk:
Hi there! Thank you for your work here.

Unfortunately I will not accept this PR — I think it would be considered very surprising that setting `width` would cause the output to change from an HTML fragment to an entire HTML document with added CSS.

The existing `width` option is documented as such:

https://github.com/kivikakk/comrak/blob/315a25c76d5547b137c1cf09135f633aa46ebf65/src/parser/mod.rs#L488

In general, styling of the result is left to the end-user, since everyone will want something different. Someone else might want or assume that `width` is measured in px instead of %. Another might not want the `margin: auto` centring.

Comrak's scope is converting CommonMark to a given output format; things that are _completely_ outside of that scope should be left out of it. Seeing as there is no provision in CommonMark itself to e.g. add a document title (such as would appear in a `<head><title>…</title></head>`), or produce output that is horizontally-centred and filling a certain width of the display, Comrak itself shouldn't try to get involved with those ideas.

In other words, if you want such a header and footer, and if you want to apply a `width: x%` CSS rule to the body, you should do that after receiving the output from Comrak, in your own script/library/program.

(The existing width option only applies to CommonMark output since there's no other way to reliably wrap such output without involving a CommonMark processor. If we ported [upstream](https://github.com/github/cmark-gfm)'s plaintext/LaTeX/manpage modes, they'd also honour it.)

Here's a shell script which produces output like you want using Comrak. The benefit is that you don't need a new release every time you want to change the styles.

```sh
#!/usr/bin/env sh

body_tag="<body>"
if test -n "$WIDTH"; then
  body_tag="<body style=\"width: $WIDTH%; margin: auto;\">"
fi

cat <<HTML
<!DOCTYPE html>
<html>
<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1">
</head>
$body_tag
$(comrak "$@")
</body>
</html>
HTML
```

You can invoke it like Comrak regularly, except if the `WIDTH` environment variable is set, it'll adjust the `<body>` tag's style accordingly.

Its-Just-Nans:
I thought after making the PR
100 % understandable

Thanks for the very detailled answer !

PR #363, opened 2024-01-31
Make non public nodes public

mfontanini:
`NodeMultilineBlockQuote` and `NodeShortCode` are not exported anywhere in this crate.

### DIFF
--- a/src/nodes.rs
+++ b/src/nodes.rs
@@ -5,7 +5,7 @@ use std::cell::RefCell;
 use std::convert::TryFrom;
 
 #[cfg(feature = "shortcodes")]
-use crate::parser::shortcodes::NodeShortCode;
+pub use crate::parser::shortcodes::NodeShortCode;
 
 pub use crate::parser::multiline_block_quote::NodeMultilineBlockQuote;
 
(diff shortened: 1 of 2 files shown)


digitalmoksha:
Hmm, it is public in `src/parser/multiline_block_quote.rs` - is that not enough? I see that `NodeShortCode` is handled in the same way.

mfontanini:
Right, the problem is that the `parser` module [is not public ](https://github.com/kivikakk/comrak/blob/main/src/lib.rs#L89). So you can have `pub` types in there but nothing is making them reachable publicly. I think it's fine that `parser` isn't `pub`, but any nodes defined within that module should be re-exported as `pub` in `comrak::nodes` like every other node type. You can see these aren't public in [the docs](https://docs.rs/comrak/0.21.0/comrak/nodes/enum.NodeValue.html#variant.ShortCode), there's no way to see their definitions there.

digitalmoksha:
Ah yeah, I see 👍

kivikakk:
LGTM! There's a little CI error because this now exposes some un-documented stuff; would you be able to address? Merge will happen automatically on CI success.

PR #367, opened 2024-03-26
Add render option to wrap escaped chars in span

digitalmoksha:
CommonMark supports [escaping characters](https://github.github.com/gfm/#backslash-escapes), which short-circuits normal markdown processing for that character.

But normally this information is lost in the final html. Post-processing filters sometimes need this information in order to keep special references from being recognized, such as `@user`. Ideally `\@user` should keep the reference from being recognized, but the final html only has `@user`.

This PR adds an option `--escaped-char-spans` that will wrap escaped characters in a `<span data-escaped-char>`. Post-processing filters can then use this when parsing for special references.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,9 @@ Options:
       --escape
           Escape raw HTML instead of clobbering it
 
+      --escaped-char-spans
+          Wrap escaped characters in span tags
+
   -e, --extension <EXTENSION>
           Specify extension name(s) to use
           
--- a/fuzz/fuzz_targets/all_options.rs
+++ b/fuzz/fuzz_targets/all_options.rs
@@ -37,6 +37,7 @@ fuzz_target!(|s: &str| {
     render.escape = true;
     render.list_style = ListStyleType::Star;
     render.sourcepos = true;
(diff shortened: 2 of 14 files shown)


kivikakk (approved):
Very very good, and super pleased to see things like sourcepos 100% correct. Thanks so much!

kivikakk:
As an aside, I can't believe I didn't think of doing something like this back when I was at GitHub and contending with the exact same problem (then in cmark-gfm) — despite very much wanting to make backslash-escaping of @mentions a possibility, I simply couldn't think of a solution, and so `\@kivikakk` still renders as @kivikakk today.

digitalmoksha:
Thanks! Yeah finally figured out a solution a couple years ago for GitLab, using [pre-processing](https://gitlab.com/gitlab-org/gitlab/blob/master/lib/banzai/filter/markdown_pre_escape_filter.rb) and [post-processing](https://gitlab.com/gitlab-org/gitlab/blob/master/lib/banzai/filter/markdown_post_escape_filter.rb) filters, but it's a hack. So I'm quite stoked to be able to add this support natively!

I just need to add some tests...

Issue #369, opened 2024-03-31
Adding Node.traverse() examples and/or documentation.


kaesluder:
I have the following example of a function using Node.traverse() to extract text as an alternative to the recursive examples from the README. If there's interest, I can clean it up and put in a PR adding it to examples or the generated API documentation.

```rust 
fn extract_text_traverse<'a>(root: &'a AstNode<'a>) -> String {
    let mut output_text = String::new();

    // Use `traverse` to get an iterator of `NodeEdge` and process each.
    for edge in root.traverse() {
        if let NodeEdge::Start(node) = edge {
            // Handle the Start edge to process the node's value.
            if let NodeValue::Text(ref text) = node.data.borrow().value {
                // If the node is a text node, append its text to `output_text`.
                output_text.push_str(text);
            }
        }
    }

    output_text
}

    #[test]
    fn extract_text_traverse_test() {
        let markdown_input = "Hello, *worl[d](https://example.com/)*";
        let arena = Arena::new();
        let options = ComrakOptions::default();
        let root = parse_document(&arena, markdown_input, &options);
        assert_eq!("Hello, world", extract_text_traverse(root));
    }

```

It should handle nested inline markup to generate a plain text string.

kivikakk:
Thank you so much, that'd be lovely! This is a really nice example!

kaesluder:
Thanks. I'll do a bit of comment editing and make sure the proper use statements are included.

kaesluder:
And, last night I found an even easier iteration method. 🤨

```rust
fn extract_text<'a>(root: &'a AstNode<'a>) -> String {
    let mut output_text = String::new();

    // Use `traverse` to get an iterator of `NodeEdge` and process each.
    for node in root.descendants() {
        if let NodeValue::Text(ref text) = node.data.borrow().value {
            // If the node is a text node, append its text to `output_text`.
            output_text.push_str(text);
        }
    }

    output_text
}
```

I might submit a dockstring PR because `descendants()` and `traverse()` have different signatures but nearly identical documentation text. (New at this.)

kaesluder:
Happy to tackle any other documentation needs.

kivikakk:
[...]

PR #370, opened 2024-04-02
add traverse() demo example

kaesluder:
Add example of using traverse() to iterate over all descendant nodes in AST. Example code strips formatting from text. 

improves #369


kivikakk:
Great; thank you so much!

PR #373, opened 2024-04-06
Avoid backslashes before a new block.

### LINKED ISSUE #372: Rendering of hard line breaks
`cmark-gfm` renders hard line breaks as two spaces followed by a newline. Since #214, `comrak` renders hard line breaks as a backslash followed by a newline. This change makes it a bit tricky to build up `comrak` ASTs programmatically, because if you put a hard line break at the end of a block it will get rendered as a backslash in the markdown. And then because commonmark specifies that a backslash at the end of a block is treated as a literal backslash instead of a line break, you will get a backslash in the output.

I tried to make a PR to change this behavior, but every option has some problems. Here are some possibilities:

1. Keep the current behavior, and require anyone constructing their own AST to be careful about inserting line breaks. This is doable, but tricky to do correctly. (For anyone running into a similar issue, our [workaround](https://github.com/tweag/nickel/pull/1880) was to use paragraph nodes instead of line breaks.)
2. Revert to the `cmark-gfm` behavior. This has the advantage of sticking close to upstream, but the disadvantage that hard line breaks at the beginning of a block no longer round-trip. This lack of a round-trip is how cmark-gfm currently behaves:
    ```
    $ echo '> \\n> a' | cmark-gfm -t commonmark test2.md | hexdump -C
    00000000  3e 20 20 20 0a 3e 20 61  0a                       |>   .> a.|
    $ echo '> \\n> a' | cmark-gfm -t commonmark test2.md | cmark-gfm -t commonmark | hexdump -C
    00000000  3e 20 61 0a                                       |> a.|
    ```
3. Make `NodeValue::LineBreak` have two variants (space-space or backslash). Downstream users get to choose which one they want. The parser remembers which one was used, and re-renders it when outputting commonmark.

What do you think? I prefer option 3, but if you're really serious about being bug-for-bug compatible with cmark-gfm then maybe 2 is better :shrug:

(Edit: I'd be happy to make a PR for either option 2 or 3. Obviously option 1 doesn't require any changes...)

jneem:
This changes the behaviour of the commonmark renderer to avoid rendering newlines as backslashes (1) at the end of a block and (2) when the next node in the current block is a block.

Fixes #372.

This isn't quite the solution that I described in the [comment](https://github.com/kivikakk/comrak/issues/372#issuecomment-2036177787), because I discovered another situation where backslash-newlines are treated differently from the double-space variant:
```text
a
\
b
```
is parsed as a single paragraph with a line break, but the double-space version is parsed as two paragraphs with no line break. I think the backslash variant is the one most likely to round-trip correctly, so I went with that as often as possible.

### DIFF
--- a/fuzz/Cargo.lock
+++ b/fuzz/Cargo.lock
@@ -128,7 +128,7 @@ checksum = "4f0807fb6f644c83f3e4ec014fec9858c1c8b26a7db8eb5f0bde5817df9c1df7"
 
 [[package]]
 name = "comrak"
-version = "0.20.0"
+version = "0.22.0"
 dependencies = [
  "arbitrary",
  "clap",
--- a/src/cm.rs
+++ b/src/cm.rs
@@ -326,6 +326,9 @@ impl<'a, 'o> CommonMarkFormatter<'a, 'o> {
                 )
                 && self.get_in_tight_list_item(node);
         }
+        let next_is_block = node
+            .next_sibling()
+            .map_or(true, |next| next.data.borrow().value.block());
 
         match node.data.borrow().value {
             NodeValue::Document => (),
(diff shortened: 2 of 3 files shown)


kivikakk:
This looks awesome; thank you so much! 🤍

If either of @digitalmoksha or @gjtorikian are able to give it a once over, I would much appreciate it! But it looks like a very nice and straightforward improvement.

kivikakk (approved):

PR #375, opened 2024-04-14
Expand traverse and descendants documentation: Issue #369

kaesluder:
original traverse documentation is a shortened version of descendants
documentation. Edited documentation comments to clarify the following:

1. traverse and reverse_traverse return an iterator of `NodeEdge`
objects. `NodeEdge` objects mark start (similar to `<div>`) and end
(similar to `</div>`) of each node.

2. descendants returns an iterator of `Node` objects. This is probably
more useful if you want to play with the Node contents.

3. added a short "Similar Functions" section to the inline documentation
to point users in the right direction.

Improves on issue #369


kivikakk:
Thank you!

Issue #290, opened 2023-03-29
MSRV bump in 0.17.0


silverpill:
In one of the recent commits, `#[default]` enum variants were introduced: https://github.com/kivikakk/comrak/commit/22aeda8ae21d3c506ce1de878825a7510357b873#diff-78fae44774dfd69260e925a4e3a4451540188d722d41690c98e75521a9b2c8c3R247
This feature requires Rust 1.62, but otherwise comrak library seems to be compatible with Rust 1.5~ (I'm using it with Rust 1.56).
If it is possible to delay MSRV increase a bit, I would appreciate that.

(This issue is specifically about library; the binary probably has higher MSRV already due to `clap` dependency)

kivikakk:
Hi there!

Yes, I'm happy to do this — would you be able to open the PR for the change? Otherwise I'll get to it in a bit.

kivikakk:
0.18 will land later today with #293 merged — I'll bump (and close) this issue when it's out!

kivikakk:
This is done!

silverpill:
Thanks!
One more question. I'm trying to figure out how to modify my code in line with v0.18.0 API changes (https://github.com/kivikakk/comrak/commit/a11a4d48d82f83f54224903e77bbe18d6dc71a6d). I used `Ast::new(<nodevalue>)` to replace nodes in existing document and add new ones, but now this constructor requires `an argument of type LineColumn`. How line and column should be determined?

kivikakk:
The nodes you're replacing will have the value you want — `node.data.borrow().sourcepos.start`.

silverpill:
@kivikakk What about adding new nodes? For, example, pre 0.18.0 it was possible to do the following:

```rust
let text_node_value = NodeValue::Text(mytext);
let text_node = arena.alloc(AstNode::from(text_node_value));
node.append(text_node);
```

In 0.18.0, `AstNode` doesn't implement `From<NodeValue>`. One can create it with `Node::new(RefCell::new(Ast::new(value, ?)))` but it is not clear what should be the starting position.

kivikakk:
Hmm, yeah, it's not too ergonomic to be forced to come up with a value for that, is it? I think I might add a `Default` implementation that Just Works:tm: to make this easier. For now try `(0, 1).into()`.

silverpill:
[...]

PR #377, opened 2024-04-18
Feat/inplace: add new parameter `--inplace` (`-i`) for in-place formatting

bioinformatist:
Some clippy lints also have been fixed separately 😺

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,6 +69,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
+name = "bitflags"
+version = "2.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


kivikakk:
Thank you, this looks excellent.

I really appreciate the extra clippy fixes! I know in the past that some of these changes caused issues for people because it caused an increase in the MSRV (see https://github.com/kivikakk/comrak/issues/290), but I think it's been long enough that I'm happy to move on. Still, before I merge, I'm going to compare the lib and bin MSRVs before and after this change, just for my own reference, and make an appropriate declaration in Cargo.toml.

Thank you again! :white_heart:

digitalmoksha:
@bioinformatist just curious, is there any difference between 

```
cargo run -- --inplace test.md
```

and

```
cargo run -- --to commonmark --output test.md test.md
```

It seems to work the same for me 🤔

kivikakk:
It is the same, and it should be! And it's fine for it to be 👍  We could expand `-i` to allow specifying multiple files in future, or directories.

digitalmoksha:
Sounds good to me. One thought I had was if the option was set we might be able to just set the proper `to` and `output` options internally.  But then we wouldn't be able to easily expand into multiple files, which would be desirable.

Issue #379, opened 2024-04-23
Change autolink to allow more protocols, same as normal markdown links


digitalmoksha:
Currently when using the `--unsafe` option with `comrak`, normal markdown links allow all schemes except for `javascript:`, `vbscript:`, `file:`, and `data:`. So `[smb](smb:///Volumes/shared/foo.pdf)` generates `<p><a href="smb:///Volumes/shared/foo.pdf">smb</a></p>`

I would like autolinks to be able to allow the same schemes, using a new option `--unsafe-autolinks`

An example usage is in GitLab, where we filter to allow autolinking `smb:` links. So `smb:///Volumes/shared/foo.pdf` would autolink.

An `--unsafe-autolinks` option would allow the `comrak` user to filter all links based on the schemes they support.

gjtorikian:
Would an additional option be necessary? Would the user of just `—unsafe` not expect these restrictions removed? Or does that go against the spec? Just trying to figure out whether another option is necessary.

digitalmoksha:
Hmm, I was worried about exposing something that would make it "more unsafe", but if you're already using `--unsafe`, then you're already getting normal markdown links with those types of urls, and the `<smb:///Volumes/shared/foo.pdf>` syntax works as well.

Their spec does say

> An [extended url autolink](https://github.github.com/gfm/#extended-url-autolink) will be recognised when one of the schemes `http://`, or `https://`, followed by a [valid domain](https://github.github.com/gfm/#valid-domain), then zero or more non-space non-< characters according to [extended autolink path validation](https://github.github.com/gfm/#extended-autolink-path-validation)

and 

> An [extended protocol autolink](https://github.github.com/gfm/#extended-protocol-autolink) will be recognised when a protocol is recognised within any text node. Valid protocols are:
>
> - mailto:
> - xmpp:

So obviously it breaks that. And `cmark-gfm -e autolink --unsafe` with `smb:///Volumes/shared/foo.pdf` as input doesn't recognize it as a link.

The spec doesn't say anything about `unsafe`, but the help says `Render raw HTML and dangerous URLs`.

So I think you're right, we probably don't need the extra option.

kivikakk:
[...]

PR #380, opened 2024-04-24
Change `relaxed-autolinks` to allow any url scheme

digitalmoksha:
This change makes the `relaxed-autolinks` option allow any url scheme to be detected.

For example, `smb://example.com` would be detected as an autolink.

Closes https://github.com/kivikakk/comrak/issues/379


kivikakk:
I've had a think about and a play with this — all LGTM. Thank you!

PR #383, opened 2024-04-27
Add  iterative search/replace example to examples and README.md

kaesluder:
revise #369 

Add iterative string search/replace example to examples and README.md. The new example uses the `AstNode.descendants()` iterator to examine all nodes in a tree. This should be easy to understand for newcomers. Code example is wrapped as a function which should be easy to copy/paste into a sandbox, file, or other setup.

Note: example would need additional error checking for production code. 

Basic "happy path" unit test in examples/iterator-replace.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,9 +187,9 @@ fn replace_text(document: &str, orig_string: &str, replacement: &str) -> String 
 }
 
 fn main() {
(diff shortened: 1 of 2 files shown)


kivikakk:
This is perfect — thank you so much, again! :blush::white_heart:

PR #413, opened 2024-05-16
Autolink should ignore wikilinks

digitalmoksha:
When the `autolink` extension is used, text nodes inside a WikiLink node should be ignored, the same way they are for links and images.


kivikakk (approved):
Ta!

Issue #371, opened 2024-04-03
Invalid rendering of a list followed by another block


yannham:
We're using comrak to build a markdown document programmatically and output it to a file (as markdown). At some point, we produce a list of items, where each item is a code snippet contained in a `NodeValue::Code`. 

The `NodeValue::Code` nodes are directly added to the list item's children, which are themselves added to the `NodeValue::List` node. Then some other content is appended.

When rendered with `format_markdown`, comrak doesn't properly insert a new line as a separation between the list and the next element. If the next element turns out to be e.g. a `Paragraph`, then it is rendered as:

```markdown
- `foo : Number`
- `foo | Even`
rest
```

instead of 

```markdown
- `foo : Number`
- `foo | Even`

rest
```

This is an issue, because the first one is different semantically, being equivalent under the commonmark spec to:

```markdown
- `foo : Number`
- `foo | Even` rest
```

Here is a debug print (in comrak 0.17.0) of the AST being invalidly rendered:

[test2.log](https://github.com/kivikakk/comrak/files/14856239/test2.log)

Looking at the rendering code in `cm.rs`, I found that strange bit: https://github.com/kivikakk/comrak/blob/26ad7543a512b1b07f7efcbcc49aca96f8a9c0ae/src/cm.rs#L407-L421

It turns out the renderer only inserts a newline after a list when the next node is code or another list. I don't really see why: we should _always_ insert a new line I believe (excepted when the list is the last element of the children of a block).

However, the bug above doesn't occur when parsing and re-emitting markdown from a source file, e.g. with the `comrak` binary. After some investigation, it seemss that `comrak` always parse list items as `NodeValue::Paragraph`. `Paragraph` does insert a new line after itself: https://github.com/kivikakk/comrak/blob/26ad7543a512b1b07f7efcbcc49aca96f8a9c0ae/src/cm.rs#L564

However, when the list is tight, I think this bunch of code cancels out the newline inserted by paragraphs: https://github.com/kivikakk/comrak/blob/26ad7543a512b1b07f7efcbcc49aca96f8a9c0ae/src/cm.rs#L100. I'm not 100% sure, but I feel like for some reason the last item _doesn't_ (although the `in_tight_list_item` doesn't seem to depend on the position of the item in the list). Or some other mechanism which let the `self.need_cr` to `2` and properly insert the required line.

[...]

PR #425, opened 2024-07-07
AST validation

yannham:
This draft PR implements a simple validation procedure for CommonMark ASTs, following the discussions in #371. The idea is to perform the additional checks that the nature of a block and the nature of its children (or the absence thereof) match, which isn't currently possible to enforce (or at least not in a agreable way) using static types.

@kivikakk The logic is pretty simple, but the tests seem to loop forever, so it sounds like there's an infinite loop somewhere. However the function is a pretty dumb tree traversal, so I wonder if I'm missing something (like `children()` is actually returning descendants, but given the API and the existence of `descendants()`, this doesn't seem to be the case).

### DIFF
--- a/src/cm.rs
+++ b/src/cm.rs
@@ -20,6 +20,14 @@ pub fn format_document<'a>(
     options: &Options,
     output: &mut dyn Write,
 ) -> io::Result<()> {
+    // Formatting an ill-formed AST might lead to invalid output. However, we don't want to pay for
+    // validation in normal workflow. As a middleground, we validate the AST in debug builds. See
+    // https://github.com/kivikakk/comrak/issues/371.
+    #[cfg(debug_assertions)]
+    root.validate().unwrap_or_else(|e| {
+        panic!("The document to format is ill-formed: {:?}", e);
+    });
+
     format_document_with_plugins(root, options, output, &Plugins::default())
 }
 
--- a/src/nodes.rs
+++ b/src/nodes.rs
@@ -666,8 +666,6 @@ impl<'a> From<Ast> for AstNode<'a> {
 /// Validation errors produced by [Node::validate].
 #[derive(Debug, Clone)]
 pub enum ValidationError<'a> {
-    /// Children were found for a node that isn't supposed to have chilrden.
-    UnexpectedChildren(&'a AstNode<'a>),
     /// The type of a child node is not allowed in the parent node. This can happen when an inline
     /// node is found in a block container, a block is found in an inline node, etc.
     InvalidChildType {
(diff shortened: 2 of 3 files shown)


kivikakk:
I think there's quite a few places in the new `validate` function where `self` should be `node`; the infinite loop happens where we `stack.extend(self.children())` on every iteration, thus making it inexhaustible.

kivikakk:
After that, I think you'll find most checks raised by the tests are of the `InlineNodeWithChildren` kind, but it's not true that inline nodes can't have children — they certainly can (and most do). (Basic examples of each: `Emph` and `Strong` are inlines, and they can contain `Text`, other `Emph` and `Strong`, `Link`, most (all?) other inlines. `Text` is an inline which cannot contain any other node.) `nodes::can_contain_type` should enumerate all the valid nestings, and should be used by the `validate` function.

The `commonmark_avoids_spurious_backslash` test constructs an invalid AST (that `LineBreak` should go on the end of `p1`, not as a child of the document) and is correctly flagged!

yannham:
> I think there's quite a few places in the new validate function where self should be node; the infinite loop happens where we stack.extend(self.children()) on every iteration, thus making it inexhaustible.

Oh, right. I started with a recursive function but it blew the stack and there's a specific test case to avoid that, so I switched to an iterative version which introduced `node` but I must have forgotten to change some `self` to `node`. Thanks

yannham:
It seems that a few dozen remaining tests still don't pass. Most of them have a `Document` node with `Text` or variants (like `SpoileredText`) as direct children, but this isn't allowed according to `can_contain_type` and the documentation (and the markdown spec I believe, which agrees that documents should only have blocks as children: https://github.com/commonmark/commonmark-spec/blob/9103e341a973013013bb1a80e13567007c5cef6f/CommonMark.dtd#L8).

What do you think? Should we relax `can_contain_type`, or should we fix the fact that such ASTs are ever produced?

yannham:
For the record, failing tests are:

```
failures:
    tests::commonmark::commonmark_avoids_spurious_backslash
    tests::footnotes::footnote_in_table
    tests::footnotes::footnote_with_superscript
    tests::spoiler::mismatched_spoilers
    tests::spoiler::spoiler
    tests::spoiler::spoiler_in_table
    tests::spoiler::spoiler_regressions
    tests::strikethrough::strikethrough
    tests::superscript::superscript
    tests::underline::underline
```

kivikakk:
Nice, thank you so much!

Of the remaining, some of them are legit (like `FootnoteReference` in a `TableCell`) and need `can_contain_type` to be adjusted, and some of them (like `Text` in `Document`) aren't. If it's all good with you, I'll make the remaining adjustments myself?

yannham:
Sure!

kivikakk:
Thanks! The spoiler ones were particularly fun: half of the issues raised were legit, but half unearthed a weirdness in what gets added to the AST when a spoiler doesn't fully match. I've left a hack in `can_contain_type` for that for now since fixing _that_ is out of scope for this PR. If both you and CI like where it's at, it should be good to go!

yannham:
I did a bit of clean up, and simplified further the validation function: the first test that some nodes don't have children at all was superfluous, because it's also subsumed by `can_contain_node` which would return false on such an unexpected child. So now, `validate` mostly just maps `can_contain_child` on each pair (child, parent) of the tree.

It's good to go on my end :slightly_smiling_face:

kivikakk:
Really clean now. Thank you so much!

kivikakk:
This just unearthed [some AST mistakes I was making](https://github.com/kivikakk/kivikakk/commit/2f99e2fd33a90588febdb1865e41f62ecf285fe0#diff-42cb6807ad74b3e201c5a7ca98b911c5fa08380e942be6e4ac5807f8377f87fcL23)! Thanks again! :)

PR #428, opened 2024-07-10
Support both upstream CommonMark and GFM's differences in the base spec.

### LINKED ISSUE #329: Need link to spec in readme + document deviations
Readme has `spec 671/671` as a badge on top, but it doesn't mention anywhere **what** spec that is.

I've noticed, that cmark-gfm changed their `spec.txt` recently (to the point that it's not commonmark anymore):
https://github.com/github/cmark-gfm/commit/828322d1ee4facdab56f0d3edccb13e9af90dcd2

But the actual specification published by github doesn't have those:
https://github.github.com/gfm/

This creates confusion about what spec is being used. Thus, I'm asking that `spec 671/671` badge would link to spec currently followed by the parser.

Also, "100% CommonMark-compatible" statement might no longer be true (`****because of this****`). Maybe document what changes from it are present, and whether it's possible to engage commonmark-compatible mode?

kivikakk:
Fixes #329.

~~WIP; I think there's some CommonMark revisions to catch up on.~~

* [x] Example 28 (lines 690-700) Entity and numeric character references
* [x] Example 354 (lines 6347-6357) Emphasis and strong emphasis
* [x] Example 540 (lines 8129-8135) Links

This brings us to CommonMark 0.31.2.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "vendor/cmark-gfm"]
 	path = vendor/cmark-gfm
-	url = https://github.com/github/cmark-gfm.git
+	url = https://github.com/kivikakk/cmark-gfm.git
 [submodule "vendor/pulldown-cmark"]
 	path = vendor/pulldown-cmark
 	url = https://github.com/raphlinus/pulldown-cmark.git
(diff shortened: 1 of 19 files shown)


digitalmoksha:
Whoa, cool! 🚀

gjtorikian:
Thank you! :)

PR #439, opened 2024-07-12
Inline sourcepos fixes.

### LINKED ISSUE #301: Unexpected sourcepos.
Hi. "sourcepos" is exactly what I was looking for.
What do you think of this result?

markdown

```md
[AB
CD](/)
```

my expected

```html
<p data-sourcepos="1:1-2:6"><a data-sourcepos="1:1-2:6" href="/">AB
CD</a></p>
```

comrak v0.18.0

```html
<p data-sourcepos="1:1-2:6"><a data-sourcepos="2:1-2:6" href="/">AB
CD</a></p>
```

kivikakk:
Fixes #301.

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -725,33 +725,31 @@ impl<'o, 'c: 'o> HtmlFormatter<'o, 'c> {
                 }
             }
             NodeValue::Text(ref literal) => {
+                // No sourcepos.
                 if entering {
                     self.escape(literal.as_bytes())?;
                 }
             }
             NodeValue::LineBreak => {
+                // No sourcepos.
                 if entering {
-                    self.output.write_all(b"<br")?;
-                    self.render_sourcepos(node)?;
-                    self.output.write_all(b" />\n")?;
+                    self.output.write_all(b"<br />\n")?;
                 }
             }
             NodeValue::SoftBreak => {
+                // No sourcepos.
                 if entering {
                     if self.options.render.hardbreaks {
-                        self.output.write_all(b"<br")?;
-                        self.render_sourcepos(node)?;
-                        self.output.write_all(b" />\n")?;
+                        self.output.write_all(b"<br />\n")?;
                     } else {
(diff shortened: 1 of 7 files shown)


kivikakk:
How awkward. In all these cases:

```markdown
- A
  B
```

```markdown
- A
B
```

```markdown
- A
   B
```

the inline parser is initialised with the subject `"A\nB"` and a `block_offset` of 2, and doesn't have any distinguishing information. This is a consequence of cmark/Comrak's particular way of collecting block content before handing it off to the inline parser — it normalises things such that the inline parser doesn't need to worry about the minutae of what falls into a block and what doesn't, but we also lose information which we cannot recover.

This will require something of a rearchitecting — sourcepos info needs to be associated and carried through from _much_ earlier in the block processing stage, with the possibility of multiple spans per content block[1] — and I have no real inclination to do such work at present.

In light of all this, for now I'm going to adjust to match upstream (https://github.com/kivikakk/comrak/issues/301#issuecomment-1539424419) and not output inline-level sourcepos in HTML in the same cases as cmark. Block-level sourcepos will be unaffected, as it's not known to have issues.

---

[1] In the examples above, we still need to be passing through the same inline `Subject::input` of `"A\nB"`, but with sourcepos marker collections `[(1:3-1:4),(2:3-2:3)]`, `[(1:3-1:4),(2:1-2:1)]` and `[(1:3-1:4),(2:4-2:4)]` respectively (where `(1:4)` aligns with the `\n` which is turned into a soft break). Right now we only pass a single point `(1:3)` and infer the rest incrementally.

This changes a lot of the bookkeeping on the block parser side, and a complete rework of sourcepos handling on the inline parser side, so I can't even say for sure that I won't hit an even bigger roadblock. Also worth noting in conjunction is [this comment about autolink sourcepos unsoundness](https://github.com/kivikakk/comrak/blob/3bb6d4ceb82e7c875d777899fc8c093d0d3e1fce/src/tests/autolink.rs#L202-L226).

PR #453, opened 2024-08-13
Track line offsets for better accuracy of inline sourcepos

digitalmoksha:
Adds a `line_offsets` vector used to record the number of spaces stripped from each line of content added. This allows us to more accurately calculate `sourcepos` for inline elements.

### DIFF
--- a/src/nodes.rs
+++ b/src/nodes.rs
@@ -534,6 +534,7 @@ pub struct Ast {
     pub(crate) open: bool,
     pub(crate) last_line_blank: bool,
     pub(crate) table_visited: bool,
+    pub(crate) line_offsets: Vec<usize>,
 }
 
 /// Represents the position in the source Markdown this node was rendered from.
@@ -609,6 +610,7 @@ impl Ast {
             open: true,
             last_line_blank: false,
             table_visited: false,
+            line_offsets: Vec::with_capacity(0),
         }
     }
 }
--- a/src/parser/inlines.rs
+++ b/src/parser/inlines.rs
@@ -31,8 +31,8 @@ pub struct Subject<'a: 'd, 'r, 'o, 'c, 'd, 'i> {
     pub input: &'i [u8],
     line: usize,
     pub pos: usize,
-    block_offset: usize,
     column_offset: isize,
+    line_offset: usize,
     flags: Flags,
     pub refmap: &'r mut RefMap,
     delimiter_arena: &'d Arena<Delimiter<'a, 'd>>,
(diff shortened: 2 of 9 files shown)


digitalmoksha:
Initial spaghetti against the wall. Try using a line offset vector.

digitalmoksha:
@kivikakk would you be able to take a look at this when you have a moment?  I think this solves one of the main issues of inline sourcepos.  Or at least gets us much closer.

I was able to solve the problems that we had with tables.  But I haven't tackled any of the autolinking problems yet.

src/nodes.rs
```diff
@@ -609,6 +610,7 @@ impl Ast {
             open: true,
             last_line_blank: false,
             table_visited: false,
+            line_offsets: Vec::with_capacity(0),
```
digitalmoksha, line 613: I changed to using `with_capacity(0)` in the second commit. The idea being there many ast nodes that will never use `line_offsets`, so why keep around default allocated storage for them.

Maybe it's not an issue. Or maybe it would be better to make it an Option that gets allocated only when we need to add something it. 🤔

src/parser/inlines.rs
```diff
@@ -182,6 +181,11 @@ impl<'a, 'r, 'o, 'c, 'd, 'i> Subject<'a, 'r, 'o, 'c, 'd, 'i> {
             None => return false,
             Some(ch) => *ch as char,
         };
+
+        let node_ast = node.data.borrow();
+        let adjusted_line = self.line - node_ast.sourcepos.start.line;
+        self.line_offset = *node_ast.line_offsets.get(adjusted_line).unwrap_or(&0);
```
kivikakk, line 187: Is this `unwrap_or` needed, or is it purely speculative? If it's the latter, I vote for this:

```suggestion
        self.line_offset = node_ast.line_offsets[adjusted_line];
```

Put another way: could this mask a bug one day? Or is it actually expected that the corresponding line offset may not be found for a valid reason? I've tried this change and the unit and spec tests pass, no crashes on fuzzing, so I think we should go with it.

See also: [Contracts, Undefined Behavior, and Defensive Programming](https://bloomberg.github.io/bde-resources/pdfs/Contracts_Undefined_Behavior_and_Defensive_Programming.pdf).

kivikakk (approved):
This is so good! Well done! You made this so much easier than I imagined. :)

src/nodes.rs
digitalmoksha: I agree, I wasn't relishing the thought of using an Option. Just wasn't sure if this was Rust-ish enough. 👍

src/parser/inlines.rs
digitalmoksha: Yeah, I was being more defensive. But I think you're right, it's not needed. And if it's passing fuzzing, then I'm more confident.

I also changed it in `table.rs`

digitalmoksha:
Thanks @kivikakk, incorporated the changes!

PR #458, opened 2024-08-22
Add a render option to render the image as <figure>

JmPotato:
Though #442 discussed a more common design to introduce this kind of customization to certain renderings, it's reasonable to me to have this option as a simple and direct solution to the `<figure>` demand.

This PR adds a new render option named `figure_with_caption` to do this.


kivikakk (approved):
LGTM — thank you!

Issue #459, opened 2024-08-22
relaxed_autolinks option causes markdown links with url in title to get parsed incorrectly.


digitalmoksha:
When using the `auotlinker` extension and the `relaxed_autolinks` option, markdown links such as `[https://github.com](https://github.com)` get parsed incorrectly, generating 

```
<p>[<a href="https://github.com/%5D(https://github.com/)">https://github.com/](https://github.com/)</a></p>
```

instead of

```
<p><a href="https://github.com/">https://github.com/</a></p>
```

See https://gitlab-org.gitlab.io/ruby/gems/gitlab-glfm-markdown/?text=%5Bhttps%3A%2F%2Fgithub.com%2F%5D(https%3A%2F%2Fgithub.com%2F)

kivikakk:
Bisect shows this was introduced at bc754e62e569684956cfc8da635a3eee878a9520; i.e. upon my attempt at restoring `relaxed_autolinks` behaviour after the inline rewrite to fix up all those other autolink edgecases.

There should be a failing unit test for this to prevent a regression; exactly how to restore the behaviour depends on exactly what the desired behaviour is. (i.e. when is an autolink inside brackets good for "relaxed autolinking", vs. not?)

digitalmoksha:
ah, thanks for bisecting that!

digitalmoksha:
I've gone back and forth on this, and I don't see a good way to solve this for the general case.

It's easy enough to fix `[http://foo.com](http://foo.com)` and make sure it generates `<a href="http://foo.com">http://foo.com</a>`.

But the more general case, `[this http://and.com that and http://the.com other](http://foo.com)` is more difficult. As it stands it generates

```
<a href="http://foo.com">this <a href="http://and.com">http://and.com</a> that and <a href="http://the.com">http://the.com</a> other</a>
```

rather than

```
<a href="http://foo.com">this http://and.com that and http://the.com other</a>
```

It's problematic to have links embedded within links.

The original problem we were trying to solve was being able to support matching bracketed text and auto links so that the ending bracket is not pulled in as part of the url. For example

```
{this http://foo.com}
```

Without `--relaxed-autolinks` we get

```
<p>{this <a href="http://foo.com%7D">http://foo.com}</a></p>
```

and with it we get

```
<p>{this <a href="http://foo.com">http://foo.com</a>}</p>
```

This is the same type of support for `(` - `(this http://foo.com)` generates what you would expect, an auto linked `foo.com`.

[...]

PR #461, opened 2024-08-26
Fix edge cases for relaxed-autolink option

digitalmoksha:
Handle several edge cases with the `relaxed-autolinks` option when autolinking inside a markdown link. We now detect if it's a single url in a markdown link (`[url](url)`). We also handle if an autolink does get added to a link by removing the duplicate link when converting to HTML. 

Related issue: #459

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -838,23 +838,33 @@ impl<'o, 'c: 'o> HtmlFormatter<'o, 'c> {
             }
             NodeValue::Link(ref nl) => {
                 // Unreliable sourcepos.
-                if entering {
-                    self.output.write_all(b"<a")?;
-                    if self.options.render.experimental_inline_sourcepos {
-                        self.render_sourcepos(node)?;
+                let parent_node = node.parent();
+
+                if !self.options.parse.relaxed_autolinks
+                    || (parent_node.is_none()
+                        || !matches!(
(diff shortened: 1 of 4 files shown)


kivikakk:
This fails fuzzing. A minimised `all_options` test case is here: [minimized-from-7305067d8bb96107d864bbcb270f09d57ab62a42.md](https://github.com/user-attachments/files/16758555/minimized-from-7305067d8bb96107d864bbcb270f09d57ab62a42.md). Repro with `cargo fuzz run all_options minimized-from-7305067d8bb96107d864bbcb270f09d57ab62a42.md`.

I assume `i + link_end` may be zero.

digitalmoksha:
Thanks @kivikakk, that does seem to be it. I've pushed a change, and am running `cargo +nightly fuzz run all_options -j 6`

kivikakk:
LGTM — thanks!

digitalmoksha:
@kivikakk are you up for publishing a new release?

kivikakk:
Sure thing; will do so in the next hour or so.

kivikakk:
It's [on crates.io](https://crates.io/crates/comrak/0.28.0) :).

digitalmoksha:
Thank you! 🙇

Issue #327, opened 2023-06-27
Description lists with tight and loose syntax


digitalmoksha:
What do you think about supporting the tight and loose syntax as laid out by `jgm` in https://github.com/jgm/commonmark-hs/blob/master/commonmark-extensions/test/definition_lists.md?

This follows the style of lists in general, where a tight list doesn't wrap the content in a paragraph, and the loose list does. So this would be valid

```
one
: two
: three
```

and generate this

```
<dl>
<dt>one</dt>
<dd>two
</dd>
<dd>three
</dd>
</dl>
```

and this would be valid

```
one

: two

: three
```

and generate this

```
<dl>
<dt>one</dt>
<dd>
<p>two</p>
</dd>
<dd>
<p>three</p>
</dd>
</dl>
```

I think what is there currently is mostly a subset of that spec.  The spec adds a tight list (no blank line necessary), support for `~` as a marker, and allows for multiple definitions for each term.

kivikakk:
SGTM — I like the idea of making the tight/loose distinction more widely used, since it's a bit obscure as it is and is often a source of confusion.

PR #462, opened 2024-09-02
Enhance description lists

digitalmoksha:
Work in progress. Enhancing description lists.

- [x] Add support for multiple description definitions

      ```
      foo

      : bar
      : baz
      ```

- [x] Allow for the definition to follow after the term without a blank line

      ```
      foo
      : bar
      ```

- [x] Allow a definition list to be either "tight" or "loose", same as a normal list. Related to https://github.com/kivikakk/comrak/issues/327

  Waffling on this one. "Tight" and "loose" lists add complexity, and is not well understood by users. Most implementations I've seen do support the distinction. However not supporting it means styling the list would be easier and less complex. 🤷


src/parser/mod.rs
```diff
@@ -1816,6 +1829,50 @@ impl<'a, 'o, 'c: 'o> Parser<'a, 'o, 'c> {
...
+            // let parent = last_child.parent().unwrap();
+            // let item = parent.last_child().unwrap();
+            // 
+            // reopen_ast_nodes(item);
+            // 
+            // let details =
+            //     self.add_child(item, NodeValue::DescriptionDetails, self.first_nonspace + 1);
+            // *container = details;
+            // 
+            // true
```
digitalmoksha, line 1875: @kivikakk I was wondering if you could take a look at this.

I realized that when supporting multiple definitions, the second `DescriptionDetails` gets added as an entirely new `DescriptionItem`, rather than being added as a child to the existing `DescriptionItem`.

But both my attempts to do this resulted in the `assert!(ast.open);` in `finalize_borrowed` tripping.

Sample input

```
foo
: bar
: one
```

I thought maybe it's the shenanigans I'm doing at line 1760, the `None` case, but using the following input (with blanks lines) bypasses that and it still asserts.

```
foo

: bar

: one
```

Even with re-opening the ast node, it's not working. I'm missing some magic incantation. 🤔

digitalmoksha: I couldn't get multiple detail nodes added to a description item. But thinking about it again, I really think it's ok - it just means that there can be multiple `DescriptionItem` with just a `DescriptionDetails`, but no term. Actually I think this is fine. The HTML formatter doesn't care.

digitalmoksha:
Hey @kivikakk I think this is done now. Running fuzzing, found a couple things initially, which are now fixed. But I'll let it run another few hours.

This has extended the syntax to allow for tight lists and multiple definition items per term, such as

```
Apple
: juice
: core

Orange
: fruit
: color
```

wdyt?

kivikakk:
Hey @digitalmoksha, that sounds good to me — I'll have a proper read over of the PR tonight! Thank you for your patience with this!

kivikakk (approved):
Bravo, this looks fantastic. Feel free to merge! 🙇‍♀️

kivikakk:
Would you like me to make a release?

digitalmoksha:
Oh yeah, that would be awesome 🙇 🙇 🙇

kivikakk:
[Done](https://crates.io/crates/comrak/0.31.0)! 🤍

PR #465, opened 2024-09-06
Add option for specifying a minimum width of ordered lists

edwar4rd:
See #464 .
I added ol_width in rendering options to specify the reserved width of ordered list markers.
For example, having `ol_width = 5` give you `1.   something` and `123. something` and `1234. something`.
A single whitespace is always added, so `ol_width <= 3` have no effect.
The previous behavior is equivalent to `ol_width = 4`.

Test `tests::commonmark::commonmark_relist` is adjusted with the changes.
During testing, `self.prefix.len() - marker_width` can sometimes cause a overflow.
I'm not too sure what do `self.prefix` mean, but 

```rust
let new_len = if self.prefix.len() > marker_width {
    self.prefix.len() - marker_width
} else {
    0
};
```

seemed to work fine.

### DIFF
--- a/src/cm.rs
+++ b/src/cm.rs
@@ -473,16 +473,22 @@ impl<'a, 'o, 'c> CommonMarkFormatter<'a, 'o, 'c> {
             let list_delim = parent.delimiter;
             write!(
                 listmarker,
-                "{}{}{}",
+                "{}{} ",
                 list_number,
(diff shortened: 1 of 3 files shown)


kivikakk:
Thank you so much! Sorry about the slow response, have not had the capacity to process this yet!

PR #466, opened 2024-09-08
Use `bon` for an infallible and compile-time-checked builder

Veetaha:
Hi! This PR replaces the usage of `derive_builder` with [`bon`](https://github.com/elastio/bon). What does it give us?

- The `build()` method no longer returns a `Result<T>`, it returns `T` instead. So there are no potential panics and no need for error handling in runtime. For example:
    image
- The builder generated by `bon` validates that the same field isn't set twice. For example, with `bon` I detected this mistake in tests where `math_dollars` and `math_code` were set twice, because `bon` generated a [compile error](https://elastio.github.io/bon/blog/bon-builder-v2-1-release#better-error-messages) https://github.com/kivikakk/comrak/blob/1a33c63688dcfabc0bb5e61fc043765d20debf83/src/tests/api.rs#L62-L67
- `bon` generates a `T::builder()` method, which removes the need for importing the `TBuilder` type and using `TBuilder::default()` syntax to create the builder.
- If you ever need some more complex logic of building, or you want to avoid create a separate "parameters" struct, you can always generate a [builder from a function](https://elastio.github.io/bon/guide/overview#builder-for-a-function) or [from a method](https://elastio.github.io/bon/guide/overview#builder-for-an-associated-method) with `bon`. Builders generated this way use the same API conventions and it is even possible to switch between struct/function syntax and keep your crate API the same (see e.g. [switching from `#[derive(Builder)]` to a `#[builder]` on the method named "new"](https://elastio.github.io/bon/guide/compatibility#switching-between-derive-builder-and-builder-on-the-new-method))

I understand this is a breaking change for this crate, but I think improvements are worth it.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,9 +1,8 @@
 //! The `comrak` binary.
 
 use comrak::{
-    adapters::SyntaxHighlighterAdapter, plugins::syntect::SyntectAdapter, Arena,
-    ExtensionOptionsBuilder, ListStyleType, Options, ParseOptionsBuilder, Plugins,
-    RenderOptionsBuilder,
+    adapters::SyntaxHighlighterAdapter, plugins::syntect::SyntectAdapter, Arena, ListStyleType,
+    Options, Plugins,
 };
 use std::boxed::Box;
 use std::env;
@@ -14,6 +13,7 @@ use std::path::PathBuf;
 use std::process;
 
 use clap::{Parser, ValueEnum};
+use comrak::{ExtensionOptions, ParseOptions, RenderOptions};
 
 const EXIT_SUCCESS: i32 = 0;
 const EXIT_PARSE_CONFIG: i32 = 2;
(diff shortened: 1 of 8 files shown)


src/main.rs
```diff
@@ -247,15 +247,14 @@ fn main() -> Result<(), Box<dyn Error>> {
...
-    extension
+    let extension = ExtensionOptions::builder()
         .strikethrough(exts.contains(&Extension::Strikethrough) || cli.gfm)
         .tagfilter(exts.contains(&Extension::Tagfilter) || cli.gfm)
         .table(exts.contains(&Extension::Table) || cli.gfm)
         .autolink(exts.contains(&Extension::Autolink) || cli.gfm)
         .tasklist(exts.contains(&Extension::Tasklist) || cli.gfm)
         .superscript(exts.contains(&Extension::Superscript))
-        .header_ids(cli.header_ids)
+        .maybe_header_ids(cli.header_ids)
```
Veetaha, line 257: For optional fields (or fields with `#[builder(default)]`) `bon` generates two setters:
- `<field>(value: T)`
- `maybe_<field>(value: Option<T>)`

This is desingned this way so that [changing a required field to optional is compatible](https://elastio.github.io/bon/guide/compatibility#making-a-required-member-optional), and it also simplifies the syntax for optional members by removing the need to wrap the value with `Some()` using the `<field>` setter

Veetaha:
**UPD**: `bon 2.2.1` added support for MSRV `1.59.0`

kivikakk:
Thanks so much for all the effort you put into this! This is amazing, it looks really nice, thank you!

I'll rebase it and get it passing CI. :)

kivikakk:
Thank you again! 🤍

PR #486, opened 2024-11-13
Prevent panic in format_item

### LINKED ISSUE #485: format_commonmark panics on a single list item
Panic occurs when `format_commonmark` input contains a single `NodeValue::Item`:

https://github.com/kivikakk/comrak/blob/v0.29.0/src/cm.rs#L468

It didn't panic before version 0.25.0 (`unwrap` was introduced in #429).

silverpill:
Fixes #485


kivikakk:
Great. Thank you so much!

PR #494, opened 2024-11-28
Account for front matter when calculating `sourcepos`

SamWilsn:
Hopefully this is sufficient to account for the front matter when calculating `sourcepos` for subsequent nodes.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -1141,6 +1141,20 @@ impl<'a, 'o> Parser<'a, 'o> {
             &self.options.extension.front_matter_delimiter,
         ) {
             if let Some((front_matter, rest)) = split_off_front_matter(s, delimiter) {
+                let lines = front_matter
+                    .as_bytes()
+                    .iter()
+                    .filter(|b| **b == b'\n')
+                    .count();
+
+                let mut stripped_front_matter = front_matter.to_string();
+                strings::remove_trailing_blank_lines(&mut stripped_front_matter);
+                let stripped_lines = stripped_front_matter
+                    .as_bytes()
+                    .iter()
+                    .filter(|b| **b == b'\n')
+                    .count();
+
                 let node = self.add_child(
                     self.root,
                     NodeValue::FrontMatter(front_matter.to_string()),
@@ -1148,6 +1162,15 @@ impl<'a, 'o> Parser<'a, 'o> {
                 );
                 s = rest;
                 self.finalize(node).unwrap();
+
+                node.data.borrow_mut().sourcepos = Sourcepos {
+                    start: nodes::LineColumn { line: 1, column: 1 },
+                    end: nodes::LineColumn {
+                        line: 1 + stripped_lines,
+                        column: delimiter.chars().count(),
+                    },
+                };
+                self.line_number += lines;
             }
         }
 
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -1167,7 +1167,7 @@ impl<'a, 'o> Parser<'a, 'o> {
                     start: nodes::LineColumn { line: 1, column: 1 },
                     end: nodes::LineColumn {
                         line: 1 + stripped_lines,
-                        column: delimiter.chars().count(),
+                        column: delimiter.len(),
                     },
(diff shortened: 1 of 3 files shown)


src/parser/mod.rs
```diff
@@ -1141,13 +1141,26 @@ impl<'a, 'o> Parser<'a, 'o> {
...
                     NodeValue::FrontMatter(front_matter.to_string()),
                     1,
                 );
                 s = rest;
                 self.finalize(node).unwrap();
+                node.data.borrow_mut().sourcepos = Sourcepos {
+                    start: nodes::LineColumn { line: 1, column: 1 },
+                    end: nodes::LineColumn {
+                        line: lines,
+                        column: delimiter.chars().count(),
```
SamWilsn, line 1160: This line I'm particularly unsure about. Are columns supposed to be bytes, characters, or actual columns (i.e. accounting for ZWJ, combining diacritics, etc.)?

SamWilsn:
Narrator: It was not sufficient.

src/parser/mod.rs
kivikakk: This question gave me pause too. Instead of me trying to infer it from some of my own code, let's try to get it from the source ([pun intended](https://www.youtube.com/watch?v=YbeZoV3Dcsc) (video unrelated)). If we ask [cmark](https://github.com/commonmark/cmark) for sourcepos of some text with minimally non-byte-oriented text …

```console
$ echo 'テスト test' | build/src/cmark --sourcepos
<p data-sourcepos="1:1-1:14">テスト test</p>
```

… we see it doesn't make any attempt to interpret UTF-8 whatsoever. This is a bit ¯\\\_(ツ)\_/¯ But it accords with our current behaviour, which makes sense given we modelled on it totally:

```console
$ echo 'テスト test' | cargo run -- --sourcepos
<p data-sourcepos="1:1-1:14">テスト test</p>
```

So, for consistency (and ease of your implementation), I'd say let's start with bytes, and I'll open an issue to think about improving this. There's a few places in `parser/mod.rs` where we actually use `self.column` and treat it as a byte count, and while I'm not sure it'd come into play here exactly (they're mostly around leading indent), it's another vote for bytes.

kivikakk (approved):
Thank you so much!

kivikakk:
I'll apply the bytes change (and test adjustment) and merge. :)

SamWilsn:
> it doesn't make any attempt to interpret UTF-8 whatsoever

I totally understand!

This is just me dreading my next step and I absolutely don't expect any effort on your behalf, but it's going to be a pain to adapt this to work with [`annotate-snippets`](https://crates.io/crates/annotate-snippets), which does do unicode interpretation :crying_cat_face:

kivikakk:
Yes, fair T_T I hope it's not too painful — let me know if I can (try to!) elucidate anything else for you about the base design.

Also, if you ever want a release made with your changes in it, just give me a ping!

SamWilsn:
Appreciate it! I have a few more pull requests to put in, so no rush on a new release.

ryanpeach:
@kivikakk if this is working, my PR https://github.com/ryanpeach/mdlinker/pull/58 would love to see a new release.

kivikakk:
Sure thing!

src/parser/mod.rs
```diff
@@ -1141,13 +1141,36 @@ impl<'a, 'o> Parser<'a, 'o> {
...
         ) {
             if let Some((front_matter, rest)) = split_off_front_matter(s, delimiter) {
+                let lines = front_matter
+                    .as_bytes()
+                    .iter()
+                    .filter(|b| **b == b'\n')
+                    .count();
+
+                let mut stripped_front_matter = front_matter.to_string();
+                strings::remove_trailing_blank_lines(&mut stripped_front_matter);
```
ryanpeach, line 1151: I'm concerned by this line. I'm experiencing an off-by-one error and there is space between my frontmatter and the beginning of my work. This would account for it.

ryanpeach: @SamWilsn

ryanpeach:
Working for me :)

PR #497, opened 2024-12-04
Make the two wikilinks extensions mutually exclusive (and other assorted improvements)

SamWilsn:
Noticed some weirdness with the commonmark formatter and having both extensions enabled (`[[title|url|title]]` kind of thing.)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,3 +78,7 @@ clap = { version = "4.0", optional = true, features = [
     "string",
     "wrap_help",
 ] }
+
+[[example]]
+name = "syntect"
+required-features = [ "syntect" ]
--- a/examples/s-expr.rs
+++ b/examples/s-expr.rs
@@ -14,7 +14,7 @@ const INDENT: usize = 4;
 const CLOSE_NEWLINE: bool = false;
 
 use comrak::nodes::{AstNode, NodeValue};
-use comrak::{parse_document, Arena, ExtensionOptions, Options};
+use comrak::{parse_document, Arena, ExtensionOptions, Options, WikiLinksMode};
 use std::env;
 use std::error::Error;
 use std::fs::File;
@@ -86,8 +86,7 @@ fn dump(source: &str) -> io::Result<()> {
         .multiline_block_quotes(true)
         .math_dollars(true)
         .math_code(true)
-        .wikilinks_title_after_pipe(true)
-        .wikilinks_title_before_pipe(true)
+        .wikilinks(WikiLinksMode::TitleFirst)
         .build();
 
     let opts = Options {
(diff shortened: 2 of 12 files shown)


digitalmoksha:
In general you're right, they should be mutually exclusive. But it seems reasonable that the behavior is undefined if you set both of them 😄 

Also, doesn't this change break backward compatibility? 🤔

Veetaha:
There is this [guide](https://bon-rs.com/guide/typestate-api/custom-methods#custom-setters) about adding custom setters, although I pointed right in the middle of it.

Veetaha:
However do we need this backwards-compat crutch though? @digitalmoksha WDYT?

SamWilsn:
You don't find that behaviour somewhat unexpected? IMO, if a weird state can be made impossible, it should be.

digitalmoksha:
Not really - it does something sensible. It's a forgiving interface. Same way in many systems if you specify an option multiple times, the last one gets used, as opposed to generating a hard error. (I checked, `comrak` generates an error if the same option is specified multiple times on the command line).

However I think it would be ok to generate an error for this. But what does that look like when using the API, where every entry point returns a `String`? I don't think panicking would be the right behavior.

Veetaha:
> I modeled this off of [Pandoc's](https://pandoc.org/MANUAL.html#extension-wikilinks_title_after_pipe) syntax. And it mentions in the [release notes](https://pandoc.org/releases.html#pandoc-3.0-2023-01-18)

Makes sense. However, `pandoc` is CLI, so it follows the CLI conventional design there. I would agree if comrak's explicit goal was to follow `pandoc`'s API 1-to-1, and be as low-level as that. However, I think that's not the case, and for comrak's higher-level Rust API I think an enum is more idiomatic in this case. 

I believe the path of least disruption would indeed be to keep the old boolean setters for backwards compat (they just set the internal enum value). Then, mark them with `#[deprecated = "use enum setter instead"]`.

Note that this will still be a breaking change for people that have this pattern in their code:

```rust
        .wikilinks_title_after_pipe(true)
        .wikilinks_title_before_pipe(true)
```

I.e. if they call both setters. This is because `bon` disallows setting the same value twice (the code just won't compile). It can however, be worked around with a [custom field](https://bon-rs.com/guide/typestate-api/custom-fields) that doesn't bear any type state. But maybe such kind breakage is actually fine since it probably points to an actual bug?

digitalmoksha:
what about

```rust
    let render_options = RenderOptions {
        wikilinks_title_after_pipe: get_bool_opt("wikilinks_title_after_pipe", options),
        wikilinks_title_before_pipe: get_bool_opt("wikilinks_title_before_pipe", options),
    };
```

digitalmoksha:
wrong code posted before. what about

```rust
    let mut comrak_options = comrak::ComrakOptions::default();

    comrak_options.extension.wikilinks_title_after_pipe = true;
    comrak_options.extension.wikilinks_title_before_pipe = true;

digitalmoksha:
@SamWilsn wondering if we should close this PR since #500 was merged?

PR #498, opened 2024-12-04
Write a sourcepos test for each `NodeValue` variant

SamWilsn:
Not sure if these are already covered elsewhere. I wrote one for my own knowledge, and kept going... A few are disabled because I believe the returned sourcepos are incorrect.

Apologies for stacking this on top of #497. Looking at just 38f24039d16af26a1b62cd42328e76e611ca45d5 might be best.

### DIFF
--- a/src/tests/sourcepos.rs
+++ b/src/tests/sourcepos.rs
@@ -386,6 +386,15 @@ const ESCAPED_TAG: TestCase = (
 after"#,
 );
 
+const ALERT: TestCase = (
+    &[sourcepos!((2:1-3:9))],
+    r#"before
+> [!note]
+> it's on
+
+after"#,
+);
+
 fn node_values() -> HashMap<NodeValueDiscriminants, TestCase> {
     use NodeValueDiscriminants::*;
 
@@ -407,6 +416,7 @@ fn node_values() -> HashMap<NodeValueDiscriminants, TestCase> {
                     | ThematicBreak // end is 4:0
                     | Link // inconsistent between link types
                     | Math // is 3:2-3:6 but should be 3:1-3:7
+                    | Raw // unparseable
             )
         })
         .filter_map(|v| {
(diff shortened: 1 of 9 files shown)


digitalmoksha:
Some of these are covered at the bottom of the specific test file, such as https://github.com/kivikakk/comrak/blob/main/src/tests/wikilinks.rs#L218-L276, though not everything is covered. So more coverage of this is super helpful.

Yeah, there is still more work to do on getting inline sourcepos working consistently.

SamWilsn:
How do you feel about the whole `strum` plus match approach to ensuring all variants are tested?

src/tests/sourcepos.rs
```diff
@@ -0,0 +1,512 @@
...
+                    | TaskItem // end is 4:0
+                    | DescriptionItem // end is 4:0
+                    | DescriptionTerm // end is 3:0
+                    | DescriptionDetails // end is 4:0
+                    | HtmlInline // end is 1:31 but should be 3:14
+                    | LineBreak // start is 1:15 but should be 1:13
+                    | Code // is 1:8-1:12 but should be 1:7-1:13
+                    | ThematicBreak // end is 4:0
+                    | Link // inconsistent between link types
+                    | Math // is 3:2-3:6 but should be 3:1-3:7
```
SamWilsn, line 409: If y'all can confirm that all of these are&mdash;in fact&mdash;bugs, I'll try my hand at fixing them.

digitalmoksha: Yeah I think those are all incorrect sourcesos 😢

kivikakk:
:wave: A lot has changed, so I've brought this PR up-to-date! I really like the approach, and I'd be happy to merge it now, rather than wait for all sourcepos to be fixed first (which will probably necessitate another big rebase).

Some of the changes in this branch have already landed in slightly different form, so a few commits got omitted; please feel free to restore to your version (or whatever you like), my force push here is mostly a suggestion in case you're a fan!

SamWilsn:
Thank you! I haven't had time to work on this in... four months apparently 😅

Go ahead with whatever commits you think are best.

kivikakk:
Thanks so much!

PR #505, opened 2024-12-12
Refactor open_new_blocks

digitalmoksha:
An attempt to refactor `open_new_blocks` a little, to help in reading the code.

@kivikakk let me know if you're ok with something like this. I wouldn't think it would affect performance too much. Let's see if our built-in benchmark has numbers.


kivikakk (approved):
Yep, happy to clean this up :)

PR #510, opened 2024-12-16
escape hash only with following whitespace

### LINKED ISSUE #509: Number/hash sign (#) is escaped
Hey, I'm using comrak to build an app on top of markdown-oxide, that uses tags `#like` `#this`;

Tweaking a source AST and writing it back to original files works great, however, tags are unnecessarily rewritten as `\#this`.

If it's about distinguishing from headers, those require a whitespace between sign and title, so this escape seems redundant, other than that I don't know why it could be needed

fominok:
this should fix #509

### DIFF
--- a/src/cm.rs
+++ b/src/cm.rs
@@ -167,7 +167,7 @@ impl<'a, 'o, 'c> CommonMarkFormatter<'a, 'o, 'c> {
                     self.begin_content = self.begin_content && isdigit(buf[i]);
                 }
             } else {
-                self.outc(buf[i], escaping, nextc);
+                self.outc(buf, i, escaping, nextc);
                 self.begin_line = false;
                 self.begin_content = self.begin_content && isdigit(buf[i]);
             }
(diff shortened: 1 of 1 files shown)


kivikakk:
Hi! Thanks for this. Unfortunately, it causes the spec tests to fail; see https://github.com/kivikakk/comrak/actions/runs/12358004998/job/34494314497?pr=510.

That's this example:

```````````````````````````````````md
Backslash-escaped `#` characters do not count as part
of the closing sequence:

```````````````````````````````` example
### foo \###
## foo #\##
# foo \#
.
<h3>foo ###</h3>
<h2>foo ###</h2>
<h1>foo #</h1>
````````````````````````````````
```````````````````````````````````

The `#` characters on the right-hand side are no longer escaped (because there's no following space), and they're treated as part of the ATX heading a result.

It looks like this may be a more complicated fix; in general, it's hard to say whether or not a character needs to be escaped to be preserved without a lot more knowledge in the formatter.

kivikakk:
#523 has been merged, which gets us a bit closer to the desired outcome here. Thanks again for your attempt! 🧡

PR #511, opened 2024-12-17
Add Raw Node

wakairo:
To address the requirements mentioned in https://github.com/kivikakk/comrak/discussions/477, Raw Node is added.

I am not sure about the code for Phase::Pre found in src/html.rs and src/xml.rs. I would greatly appreciate it if you could review them.

### DIFF
--- a/src/cm.rs
+++ b/src/cm.rs
@@ -366,6 +366,7 @@ impl<'a, 'o, 'c> CommonMarkFormatter<'a, 'o, 'c> {
             NodeValue::HtmlInline(ref literal) => {
                 self.format_html_inline(literal.as_bytes(), entering)
             }
+            NodeValue::Raw(ref literal) => self.format_raw(literal.as_bytes(), entering),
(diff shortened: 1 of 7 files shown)


kivikakk:
Thank you! The code around `Phase::Pre` in those files is related to processing alt-text — I think the codepath is not actually used in the XML version, so I should check that myself!

Your change looks great. Thanks so much. I'll review this more carefully a bit later on tonight for me, but so far it looks good and CI is passing! 🤍

kivikakk:
Thanks so much! Let me know if you'd like a release made with your new change.

wakairo:
Yes! If releasing it is straightforward, I would love to see a release with Raw Node so I can start related tasks at work. (I am also hoping that Commonmarker, the Ruby wrapper, gets a new release soon since my work is in Ruby.)

wakairo:
Thank you so much for your review and for v0.32.0!

PR #514, opened 2024-12-18
Eliminate `regex` and `once_cell` dependencies.

nicoburns:
## Objective

Reduce compile time / binary size impact of Comrak for projects that are not otherwise using the `regex` crate.

## Changes made
- Replace use of regex in a doctest with a simple `String::contains`
- Replace the usage of regex in anchorization code with `unicode_categories` crate which comrak already depends on.
- Removes `once_cell` dependendency, as it was only being used to cache a Regex

Comrak also currently also pulls in `regex` as a build dependency of `caseless`. But I have also sent that crate a PR to remove it's usage.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -165,8 +165,6 @@ dependencies = [
  "entities",
  "memchr",
  "ntest",
- "once_cell",
- "regex",
  "shell-words",
  "slug",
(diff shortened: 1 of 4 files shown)


kivikakk:
Wonderful. Thank you!

PR #515, opened 2024-12-18
Make bon builders optional

nicoburns:
## Objective

Reduce compile time / binary size impact of Comrak for projects that are not otherwise using the `regex` crate.

## Changes made
- Add `bon` feature flag
- Make bon builder derive/attributes conditional on this feature flag

## Notes

- Bon is particularly heavy on dependencies (bringing in syn, darling and more), and is not that widely used so is likely to be something that projects could eliminate from their dependency trees entirely.
- With this PR + #514 + https://github.com/unicode-rs/rust-caseless/pull/20:
   - A *release* build of Comrak with `--no-default-features` build in `1.9s`
   - The `headers` example (with default features disabled) produces a ~700kb binary

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,3 +81,7 @@ clap = { version = "4.0", optional = true, features = [
 [[example]]
 name = "syntect"
 required-features = [ "syntect" ]
+
+[[example]]
+name = "s-expr"
+required-features = [ "bon" ]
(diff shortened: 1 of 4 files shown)


kivikakk:
Thank you, again! This is a huge reduction in build-times. 🎉

Issue #472, opened 2024-10-06
Admonition Blocks extension


mhanberg:
## Description

It would be great to be able to create admonition blocks, similar to how you can on [GitHub](https://github.com/orgs/community/discussions/16925)

```markdown
> [!NOTE]  
> Highlights information that users should take into account, even when skimming.

> [!TIP]
> Optional information to help a user be more successful.

> [!IMPORTANT]  
> Crucial information necessary for users to succeed.

> [!WARNING]  
> Critical content demanding immediate user attention due to potential risks.

> [!CAUTION]
> Negative potential consequences of an action.
```

![CleanShot 2024-10-06 at 17 51 52@2x](https://github.com/user-attachments/assets/159c7b75-71a5-4856-8312-aff6ceb05566)

On GitHub is currently produces HTML that looks like

```html
<div class="markdown-alert markdown-alert-note" dir="auto">
  <p class="markdown-alert-title" dir="auto">
    <svg class="octicon octicon-info mr-2" viewBox="0 0 16 16" version="1.1" width="16" height="16" aria-hidden="true"><path d="M0 8a8 8 0 1 1 16 0A8 8 0 0 1 0 8Zm8-6.5a6.5 6.5 0 1 0 0 13 6.5 6.5 0 0 0 0-13ZM6.5 7.75A.75.75 0 0 1 7.25 7h1a.75.75 0 0 1 .75.75v2.75h.25a.75.75 0 0 1 0 1.5h-2a.75.75 0 0 1 0-1.5h.25v-2h-.25a.75.75 0 0 1-.75-.75ZM8 6a1 1 0 1 1 0-2 1 1 0 0 1 0 2Z"></path></svg>
  Note
  </p>
  <p dir="auto">Highlights information that users should take into account, even when skimming.</p>
</div>
```

Rendering an SVG is much too opinionated, but I feel that rendering a div with two p tags inside and appropriate class names to style would work well.

I am interested in contributing this feature as long as there is upstream interest, thank you!

digitalmoksha:
@mhanberg I think it would be great if you want to contribute that. It is on my list to contribute it in the next couple of months if someone else hasn't beaten me to it.

I have thought about it some, so here are a couple thoughts on the implementation.

[...]

PR #519, opened 2025-01-17
Add GitHub style alerts / admonitions

digitalmoksha:
This adds support for [GitHub style alerts](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts), such as 

```
> [!NOTE]
> Coming very soon
```

> [!NOTE]
> Coming very soon

All the types are supported, `NOTE`, `TIP`, `IMPORTANT`, `WARNING`, and `CAUTION`

A couple enhancements have been made:

1. you can use them anywhere a blockquote can be used, such as in a list item. So this will work:

   ```
   - item one

     > [!note]
     > Coming soon
   ```

2. you can override the title that's displayed by putting text after the trigger. For example:

   ```
   > [!note] Exciting news!
   > Coming soon
   ```

   Instead of showing "Note" as the title, it will show "Exciting news!"

Related to https://github.com/kivikakk/comrak/issues/472


digitalmoksha:
This is fully working. Just need to

- add an option to enable it
- add a couple more tests, including for `sourcepos`

src/xml.rs
```diff
@@ -286,6 +286,23 @@ impl<'o, 'c> XmlFormatter<'o, 'c> {
...
+                    if alert.title.is_some() {
+                        let title = alert.title.as_ref().unwrap();
+
+                        self.output.write_all(b" title=\"")?;
+                        self.escape(title.as_bytes())?;
+                        self.output.write_all(b"\"")?;
+                    }
+
+                    if alert.multiline {
+                        self.output.write_all(b" multiline=true")?;
```
kivikakk, line 303:
```suggestion
                        self.output.write_all(b" multiline=\"true\"")?;
```

kivikakk (approved):
LGTM! (Minor change: XML attribute needs to be quoted.)

src/xml.rs
digitalmoksha: Ah good catch, thanks!

digitalmoksha:
I did run fuzzing with `all_options` with no problems.  I tried `quadratic` and there was quickly a failure. But that happens on the current `main`, so it's not due to this PR.

nicoburns:
Just wanted to express my appreciation for this feature! Sometimes open source is awesome: I just bump my comrak version  (+enable the option) and suddenly I have support for this feature!

PR #521, opened 2025-01-20
Enable GitLab multiline alerts

digitalmoksha:
When the extensions `alerts` and `multiline-block-quotes` are enabled, then we now recognize GitLab style multiline alerts.  For example

```
>>> [!note]
Something of interest.

And another thing...
>>>
```

will behave the same as

```
> [!note]
> Something of interest.
>
> And another thing...
```

This makes it easier to make an alert out of multiple paragraphs and blocks, in the same fashion as [GitLab multiline block quotes](https://docs.gitlab.com/ee/user/markdown.html#multiline-blockquote)

### DIFF
--- a/script/cibuild
+++ b/script/cibuild
@@ -52,6 +52,8 @@ python3 spec_tests.py --no-normalize --spec ../../../src/tests/fixtures/descript
   || failed=1
 python3 spec_tests.py --no-normalize --spec ../../../src/tests/fixtures/alerts.md "$PROGRAM_ARG -e alerts" \
   || failed=1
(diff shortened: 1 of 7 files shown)


kivikakk:
Looking good! Have run fuzzing for a bit and will continue to do so, but I'm happy if you're happy. :)

digitalmoksha:
Yeah I feel good about. Also ran fuzzing for a couple hours 👍 

So I'll merge. Maybe we can cut a release whenever you have some time.

charlottia:
Will make one tonight or tomorrow! :)

charlottia:
Done! <https://github.com/kivikakk/comrak/releases/tag/v0.34.0>

PR #523, opened 2025-01-21
commonmark: experimental minimize

### LINKED ISSUE #509: Number/hash sign (#) is escaped

charlottia:
Fixes #509, kinda. Supersedes #510.
Experimental-quality CommonMark output minimizer.

Doesn't encode any intelligence about the underlying format; tries removing every backslash (`\`) escape in order and sees if the document parses the same, accumulating those deletions that do.

It's wasteful but better than nothing!

### DIFF
--- a/examples/iterator_replace.rs
+++ b/examples/iterator_replace.rs
@@ -27,7 +27,7 @@ fn main() {
     let doc = "This is my input.\n\n1. Also [my](#) input.\n2. Certainly *my* input.\n";
     let orig = "my";
     let repl = "your";
-    let html = replace_text(&doc, &orig, &repl);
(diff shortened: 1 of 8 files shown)


kivikakk:
Tried here, works well: https://github.com/kivikakk/kivikakk/commit/2f99e2fd33a90588febdb1865e41f62ecf285fe0.

PR #528, opened 2025-01-24
Strip trailing newlines in unhighlighted code blocks

### LINKED ISSUE #527: Unhighlighted code blocks contain invalid trailing newlines
This is the same issue as https://github.com/kivikakk/comrak/issues/278, except that it still occurs for codeblocks that are *not* syntax highlighted.

The easiest way to reproduce this is to comment out the line `plugins.render.codefence_syntax_highlighter = Some(&adapter);` in the `syntect` example. Or alternatively run the following code:

```rust
use comrak::{markdown_to_html, Options};

fn main() {
    let input = concat!("```Rust\n", "fn main<'a>();\n", "```");
    let formatted = markdown_to_html(input, &Options::default());
    println!("{}", formatted);
}
```

this prints

```no-highlight
<pre><code class="language-Rust">fn main&lt;'a&gt;();
</code></pre>
```

rather than

```no-highlight
<pre><code class="language-Rust">fn main&lt;'a&gt;();</code></pre>
```

nicoburns:
Fixes #527.

It is possible to apply a similar fix to math code blocks, but I was unsure whether that was a correct change as I'm not using math code blocks so I have not included that change here.

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -582,7 +582,10 @@ where
                         let mut code_attributes: HashMap<String, String> = HashMap::new();
                         let code_attr: String;
 
-                        let literal = &ncb.literal.as_bytes();
+                        let literal = &ncb
+                            .literal
+                            .trim_end_matches(|c| c == '\r' || c == '\n')
+                            .as_bytes();
                         let info = &ncb.info.as_bytes();
 
                         if !info.is_empty() {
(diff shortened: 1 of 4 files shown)


kivikakk:
Trailing newlines are actually per spec; see [CommonMark § 4.5](https://spec.commonmark.org/0.31.2/#fenced-code-blocks), and the spec failure in CI.

nicoburns:
Hmmph! Annoying, but it seems like you're correct (and I guess it does make some sense). I think I will need to enable highlighting to get solve this issue in my app.

PR #532, opened 2025-02-08
make `AlertType` methods public

fiji-flo:
This would allow to use `AlertType.default_title` and `AlertType.css_class` in custom hanlders and it's aligned with most other interfaces.

Shout-out to @digitalmoksha for implementing this in the first place 💯


digitalmoksha:
@fiji-flo yep, good call. Let's get this merged.

PR #533, opened 2025-02-09
Stop at first suitable $ at inline math

Bubbis:
I added two failing test cases which this change attempts to fix. 

The issue originally relates to: https://gitlab.com/gitlab-org/gitlab/-/issues/431890


digitalmoksha:
@Bubbis thanks for this fix, it looks good 🚀

PR #366, opened 2024-02-18
Add math support

digitalmoksha:
This will add support for math syntax, both the dollar math and the newer GitLab/GFM syntax. Relates to https://github.com/kivikakk/comrak/issues/23

Display math is mostly working, `$$` and ```` ```math ```` syntax. Inline not at all yet.

A couple notes:

- there is no inline math support yet
- Currently the `$$` syntax follows the rules of code blocks.  Meaning that 

  ```
  This is a paragraph
  $$
  1 + 2
  $$
  and this is a separate paragraph.
  ```

  generates a paragraph followed by a block and then another paragraph. I've seen some implementations where it is all in one paragraph. I'm torn which way it should be. At the moment this is the easiest.

- The `$$` math blocks are generated the same way that ```` ```math ```` blocks are generated, as a `<pre><code>` block. This keeps the code and the output consistent. It also means that if math rendering is not enabled on a site, the math will display as a code block, rather than a bunch of run-together text, without any extra styling needed.

- I'm using data attributes rather than classes.  So it's `<pre data-math-style="display"><code>...` rather than `<pre class="math"><code>...`.

I'm flexible on these decisions, this is just where it's at right now.

PR #534, opened 2025-02-09
Support language-specific codefence renderers

tt:
I was looking for a way to render [Pikchr](https://pikchr.org/) diagrams. I'm porting some code from Go where I used [goldmark](https://github.com/yuin/goldmark) and the [goldmark-pikchr](https://github.com/jchenry/goldmark-pikchr) extension which support this through fenced code blocks.

I thought of different ways to support this:

1. **Pikchr could be officially supported either generally or through a named feature.** This would solve my particular need but isn't particularly extensible.

2. **Implement [`SyntaxHighlighterAdapter`](https://docs.rs/comrak/0.35.0/comrak/adapters/trait.SyntaxHighlighterAdapter.html) and short-circuit the rendering of a particular language.** The problem with this is that you can't opt out of the closing `</code>` and `</pre>` tags:
  https://github.com/kivikakk/comrak/blob/8d30848b65354d67f5d1a945ddd24fdc1c0b15cb/src/html.rs#L644

3. **Pre-render certain [`CodeBlock`](https://docs.rs/comrak/0.35.0/comrak/nodes/enum.NodeValue.html#variant.CodeBlock) nodes to [`HtmlBlock`](https://docs.rs/comrak/0.35.0/comrak/nodes/enum.NodeValue.html#variant.HtmlBlock).** My one problem with this approach is that it ended up being more code than ideal and it didn't feel right.

4. **Allow a way to register language-specific renderers.** I liked this the best and there's even some prior art in #366:
  https://github.com/kivikakk/comrak/blob/8d30848b65354d67f5d1a945ddd24fdc1c0b15cb/src/html.rs#L575-L576

I would not be opposed to contributing Pikchr as a gated feature if it would be useful to others but I like adding the general capability to allow experimenting with it and other language-specific renderers.

Anyway, I opened this draft pull request so there's something concrete to discuss. If we agree that this is the preferred approach, I'll add tests, improve the documentation and address eventual feedback.

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -574,6 +574,23 @@ where
                 if entering {
                     if ncb.info.eq("math") {
                         self.render_math_code_block(node, &ncb.literal)?;
+                    } else if let Some(adapter) = self
+                        .plugins
+                        .render
+                        .codefence_renderers
+                        .get(&ncb.info)
+                    {
+                        self.cr()?;
+
+                        adapter.write(
+                            self.output,
+                            &ncb.literal,
+                            if self.options.render.sourcepos {
+                                Some(node.data.borrow().sourcepos)
+                            } else {
+                                None
+                            },
+                        )?;
                     } else {
                         self.cr()?;
 
--- a/src/adapters.rs
+++ b/src/adapters.rs
@@ -7,6 +7,17 @@ use std::io::{self, Write};
 
 use crate::nodes::Sourcepos;
 
+/// Implement this adapter for custom rendering of codefence blocks.
+pub trait CodefenceRendererAdapter: Send + Sync {
+    /// Render the codefence block.
+    fn write(
+        &self,
+        output: &mut dyn Write,
+        literal: &String,
+        sourcepos: Option<Sourcepos>,
+    ) -> io::Result<()>;
+}
+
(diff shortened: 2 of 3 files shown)


src/html.rs
```diff
@@ -574,6 +574,23 @@ where
                 if entering {
                     if ncb.info.eq("math") {
                         self.render_math_code_block(node, &ncb.literal)?;
```
tt, line 576: Arguably, this could be converted to a language-specific renderer that's registered by default.

digitalmoksha:
Hmm, interesting. The way I've done it is process the HTML using Nokogiri, and then process any code blocks that have the graph languages I support, such as mermaid, plantuml, kroki.io (which also supports pikchr), etc.

But I can see the utility of being able to have a plugin that handles registered language extensions.

kivikakk:
I'm quite a fan of this approach, and would be happy to merge it. (Likewise, refactoring the math codeblock support to use it!)

photino:
👍 for this approach.

neilberkman:
I opened a follow-up PR at #751.

kivikakk:
Thanks @neilberkman and @tt! Closing this as #751 has merged :)

Issue #538, opened 2025-02-16
[Announce] Python bindings for comrak on PyPI 🐍 💓 🦀


lmmx:
Hi there!  
 
I wanted to pop in to let you know that I packaged up some very simple Python bindings for Comrak ([Python package](https://pypi.org/project/comrak) + [repo](https://github.com/lmmx/comrak)), and saw benchmarks of 15-60x faster than the other Python packages for this task when tested with hyperfine (details in repo). It re-uses the Rust library under the hood via PyO3 in the standard way as a cdylib, so Python users can now get Comrak’s speed and CommonMark/GFM features. I thought you might like to know 😁 Do feel free to mention it in your README as a related or downstream project if you wish to signpost to others but no need at all, and thanks for all your work on Comrak, this has been really fun to piece together. I have some initial test coverage here and all seems working as planned. I think if anything I am more likely to do docs as a priority.
 
All the best :-)

Louis

digitalmoksha:
Wow, very cool!

kivikakk:
Ah, how lovely :)  Thank you! It occurs to me we don't have a bindings section at all in the README; we should definitely fix that and list this!

digitalmoksha:
#539 was merged, so I think we can close this.

MMashayekh23:
0x725f77aF07930A8312ae1d89A3A4194971335551

PR #539, opened 2025-02-21
docs: add Python bindings repo and PyPI links to Related Projects

lmmx:
Adding a note in the README about the Python bindings, as requested in #538, under _Related projects_, linking directly to the PyPI package for anyone looking to `pip install` in Python :snake:  :+1:


gjtorikian:
Thanks!

PR #540, opened 2025-02-22
Create custom HTML formatters.

kivikakk:
o/ @digitalmoksha 

See `examples/custom_formatter.rs` for what this looks like when being used from a different crate.

I've refactored some things out, added the new `html::Context` struct (which contains everything `HtmlFormatter` itself used to), moved `WriteWithLast`'s functionality into that, added `html::RenderMode` (instead of the `plain` boolean).

Still need to document newly public things!

Let me know your thoughts.


digitalmoksha:
Looking pretty good! I think we're going to need to have the context and the node available within the formatter in order to be able to do certain overrides. It doesn't look like that's exposed at the moment.

digitalmoksha:
Love it, you've really separated out the macro, it's getting pretty clean.

kivikakk:
OK! Keep in mind you may not actually always need the node itself (i.e. the `AstNode`); a lot of the time what you want will be available in the pattern-matched part.

I've updated this to now expose a variety of things, depending on the names (!) of the captures. I've also added the ability to suppress children from being rendered, so you can completely take control of a subtree. The updated example shows off all the captures available:

```rust
create_formatter!(CustomFormatter, {
    NodeValue::Emph => |output, entering| {
        if entering {
            output.write_all(b"<i>")?;
        } else {
            output.write_all(b"</i>")?;
        }
    },
    NodeValue::Strong => |context, entering| {
        use std::io::Write;
        context.write_all(if entering { b"<b>" } else { b"</b>" })?;
    },
    NodeValue::Image(ref nl) => |output, node, entering, suppress_children| {
        assert!(node.data.borrow().sourcepos == (3, 1, 3, 18).into());
        if entering {
            output.write_all(nl.url.to_uppercase().as_bytes())?;
            *suppress_children = true;
        }
    },
});
```

digitalmoksha:
Looks good. I'm frazzled for the night. I'll have a chance to look more closely on Sun. But really, I think it's about ready to go.

kivikakk:
:+1: I'll finish off the documentation; let me know, do try taking it for an actual spin (using a Git branch dependency) before we move to merge!

digitalmoksha:
I've already got one in progress 😄

src/html/context.rs
```diff
@@ -0,0 +1,88 @@
...
+pub struct Context<'o, 'c> {
+    output: &'o mut dyn Write,
+    last_was_lf: Cell<bool>,
+
+    /// [`Options`] in use in this render.
+    pub options: &'o Options<'c>,
+    /// [`Plugins`] in use in this render.
+    pub plugins: &'o Plugins<'o>,
+
+    pub(super) anchorizer: Anchorizer,
```
digitalmoksha, line 19: I think we should make this public in general. I'm trying to override `render_heading` and would need access to the `anchorizer`

src/html.rs
```diff
@@ -32,1559 +46,1730 @@ pub fn format_document_with_plugins<'a>(
...
-        if entering {
-            self.output.write_all(b"<span")?;
-            if self.options.render.experimental_inline_sourcepos {
-                self.render_sourcepos(node)?;
+    Ok(RenderMode::HTML)
+}
+
+// Helpers
+
+fn collect_text<'a>(node: &'a AstNode<'a>, output: &mut Vec<u8>) {
```
digitalmoksha, line 1544: We might consider making this public. In overriding `render_heading`, I need this function.

src/html.rs
```diff
@@ -32,1559 +46,1730 @@ pub fn format_document_with_plugins<'a>(
...
+                node: &'a $crate::nodes::AstNode<'a>,
+                entering: bool,
+            ) -> ::std::io::Result<::std::option::Option<$crate::html::RenderMode>> {
+                match node.data.borrow().value {
+                    $(
+                        $pat => {
+                            let mut suppress_children = false;
+                            $crate::formatter_captures!((context, node, entering, suppress_children), ($( $capture ),*));
+                            $case
+                            if suppress_children {
```
digitalmoksha, line 186: My first attempt at this was lifting straight out of `html.rs`, like this

```rust
    ...
    NodeValue::Heading(_) => |context, node, entering| {
        render_heading(context, node, entering);
    },

    ...
fn render_heading<'a>(
    context: &mut Context,
    node: &'a AstNode<'a>,
    entering: bool,
) -> io::Result<RenderMode> {
    let NodeValue::Heading(ref nch) = node.data.borrow().value else {
        panic!()
    };
```

In other words using what is there as an example to quickly get the right structure.

Setting aside the fact that I could pass in `nch`, the main difference is the return code. In the `render_` functions we return either `Ok(RenderMode::Plain)` or `Ok(RenderMode::HTML)`

But we're handling that automatically in the macro. So I get warnings that I'm not using the return code when I call my own `render_heading`. But the other difficulty would be if my override calls the `format_node_default`, or I want to override the `NodeValue::Image` where we do need to be able to return either `Ok(RenderMode::Plain)` or `Ok(RenderMode::HTML)`

src/html.rs
```diff
@@ -32,1559 +46,1730 @@ pub fn format_document_with_plugins<'a>(
...
-    }
+/// This must be exported so its uses in [`create_formatter!`] can be expanded,
+/// but it's not intended for direct use.
+#[doc(hidden)]
+#[macro_export]
+macro_rules! formatter_captures {
+    (($context:ident, $node:ident, $entering:ident, $suppress_children:ident), context, $bind:ident) => {
+        let $bind = $context;
+    };
+    (($context:ident, $node:ident, $entering:ident, $suppress_children:ident), output, $bind:ident) => {
```
digitalmoksha, line 208: I wonder if we need both `output` and `context`. In order to call the `cr()` function, I have to use the `context`. But I can still call `write_all` using the `context`.

But `output` doesn't have access to `cr()`, which seems like is an output function to me.

Would it be simpler just to always use `context`, and remove the `output` capture?

digitalmoksha (commented):
@kivikakk this is really good! Really appreciate you working on this and fleshing it out. I just had a couple minor comments.

I picked overriding `NodeValue::Heading` because I actually need to modify it to add `aria-describedby` to give better assistance support.

src/html.rs
kivikakk: Sure thing.

src/html.rs
kivikakk: Your code example above doesn't actually `return` at all, so I am assuming there's something missing. (Please give me reproducible examples and the actual errors in future — I don't want to have to guess what you've written, in order to reproduce the errors you've actually seen, in order to help make this suitable for you.)

But yes, the interface as I've designed it doesn't have the flexibility to choose a render mode, to simplify the main case. I'll have a think, one sec.

kivikakk: Alright, I've updated the interface; instead of using a capture (`suppress_children`), you now `return` a `comrak::html::ChildRendering` (one of `HTML`, `Plain` or `Skip`), or leave it without a `return` for the default of `HTML`.

digitalmoksha: > Your code example above doesn't actually return at all, so I am assuming there's something missing. (Please give me reproducible examples and the actual errors in future — I don't want to have to guess what you've written, in order to reproduce the errors you've actually seen, in order to help make this suitable for you.)

You're right, sorry about that. I should have pasted in the full example.

digitalmoksha:
@kivikakk this is awesome! It's working well for me. Really appreciate you knocking this out 😍 

<details><summary>Here's the code I'm using</summary>

```rust
pub fn markdown_to_html_with_plugins(md: &str, options: &Options, plugins: &Plugins) -> String {
    let arena = Arena::new();
    let root = parse_document(&arena, md, options);
    let mut bw = BufWriter::new(Vec::new());

    CustomFormatter::format_document_with_plugins(root, options, &mut bw, plugins).unwrap();
    String::from_utf8(bw.into_inner().unwrap()).unwrap()
}

create_formatter!(CustomFormatter, {
    NodeValue::Heading(_) => |context, node, entering| {
        return render_heading(context, node, entering);
    },
});

fn render_heading<'a>(
    context: &mut Context,
    node: &'a AstNode<'a>,
    entering: bool,
) -> io::Result<ChildRendering> {
    let NodeValue::Heading(ref nch) = node.data.borrow().value else {
        panic!()
    };

    match context.plugins.render.heading_adapter {
        None => {
            if entering {
                context.cr()?;
                write!(context, "<h{}", nch.level)?;
                render_sourcepos(context, node)?;
                context.write_all(b">")?;

                if let Some(ref prefix) = context.options.extension.header_ids {
                    let mut text_content = Vec::with_capacity(20);
                    collect_text(node, &mut text_content);

                    let mut id = String::from_utf8(text_content).unwrap();
                    id = context.anchorizer.anchorize(id);
                    write!(
                        context,
                        "<a href=\"#{}\" aria-hidden=\"true\" class=\"anchor\" id=\"{}{}\">",
                        id, prefix, id
                    )?;
                }
            } else {
                if let Some(ref _prefix) = context.options.extension.header_ids {
                    write!(context, "</a>")?;
                }
                writeln!(context, "</h{}>", nch.level)?;
            }
            Ok(ChildRendering::HTML)
        }
        Some(_adapter) => { format_node_default(context, node, entering) }
    }
}
```

</details>

I think this is ready to go! 🚀

kivikakk:
Awesome! I'll merge and follow up with a release sometime later today most likely.

kivikakk:
This is live in 0.36.0 ([crates.io](https://crates.io/crates/comrak/0.36.0), [GitHub release](https://github.com/kivikakk/comrak/releases/tag/v0.36.0)).

PR #548, opened 2025-03-11
Add Commonmarker to related projects

gjtorikian:
As the title says! I also:

* Alphabetized the languages (OCD)
* Moved the note about GFM implementation higher up


kivikakk:
Yippee! Thank you!

Issue #558, opened 2025-04-02
Non-trivial tasklists parsed/rendered incorrectly


Turbo87:
It looks like tasklists are currently only supported for plain text contents:

### Input

```md
- [ ] `c`
- [ ] [link](https://crates.io)
- [ ] [link](#anchor)
```

### Expected

- [ ] `c`
- [ ] [link](https://crates.io)
- [ ] [link](#anchor)

```html
<ul>
<li><input type="checkbox" disabled=""> <code>c</code></li>
<li><input type="checkbox" disabled=""> <a href="https://crates.io" rel="nofollow noopener noreferrer">link</a></li>
<li><input type="checkbox" disabled=""> <a href="#anchor" rel="nofollow noopener noreferrer">link</a></li>
</ul>
```

### Output

```html
<ul>
<li><input type="checkbox" disabled=""></li>
<li><input type="checkbox" disabled=""></li>
<li><input type="checkbox" disabled=""></li>
</ul>
```

### Related

- https://github.com/rust-lang/crates.io/pull/10913

digitalmoksha:
Hmm, indeed @Turbo87 - thanks for opening this!

I bisected it down to the commit https://github.com/kivikakk/comrak/commit/1c55e16f4fe027d9de76250a30f7c8edd906d357. It looks like the `parent.detach();` on line 3084 is the culprit. Comment says it was added to fix the fuzz test `tests::fuzz::echaw9`.

digitalmoksha:
It's only happening when something like `<code>` is at the start of the item. So ``- [ ] test `code` `` works but ``-[ ] `code` `` doesn't.

@kivikakk I haven't yet found the right combination of checks to have it do what you want. Maybe if the parent (which is a paragraph) has only one child node 🤔

Turbo87:
I can confirm that https://github.com/kivikakk/comrak/pull/559 fixes the issue. thank you @digitalmoksha! ❤

PR #559, opened 2025-04-05
Only delete parent if the node has no siblings

digitalmoksha:
This addresses https://github.com/kivikakk/comrak/issues/558. Fixes a problem where task list items that starts with an inline node other than text that does not render.

Example:

```
- [ ] `code`
```

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -3080,7 +3080,7 @@ where
 
         // See tests::fuzz::echaw9. The paragraph doesn't exist in the source,
         // so we remove it.
-        if sourcepos.end.column < adjust {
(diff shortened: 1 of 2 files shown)


digitalmoksha:
@kivikakk This should fix https://github.com/kivikakk/comrak/issues/558. We should probably cut another release.

kivikakk:
Thanks, @digitalmoksha, this looks good. I'm going to run fuzzing for a couple hours here first (just because I found all those ridiculous `echaw` test cases in tasklists through fuzzing, and the sourcepos handling stuff there was no end of failing asserts), then will release. :)

PR #580, opened 2025-06-30
Syntect syntax highlighting: Make a CSS class prefix configurable.

### LINKED ISSUE #579: Syntect syntax highlighting: specify CSS class prefix
Hello,
In my setup, I have (and want) to use CSS to style the syntax highlighting. Right now, you can instruct the Syntect syntax-highlighter to emit CSS classes by calling `css()` on the `SyntectAdapterBuilder`. Unfortunately, [the class style is hardcoded to `ClassStyle::Spaced`](https://github.com/kivikakk/comrak/blob/ff33d99ceecf2806f71e3cce182392b089787a13/src/plugins/syntect.rs#L60). In my case, this leads to collisions of CSS classes which ultimately results in weird looking syntax highlighting. An easy and proper mitigation of this issue is to allow the user to specify a prefix for the CSS classes emitted by the `SyntectAdapter`.

Since Syntect has the necessary functionality, you only have to make it configurable in the `SyntectAdapter`/`SyntectAdapterBuilder`.
I created a small PR (#580) and am looking forward to your feedback!

SteveBinary:
Fixed #579

### DIFF
--- a/src/plugins/syntect.rs
+++ b/src/plugins/syntect.rs
@@ -19,6 +19,7 @@ pub struct SyntectAdapter {
     theme: Option<String>,
     syntax_set: SyntaxSet,
     theme_set: ThemeSet,
+    css_class_prefix: Option<&'static str>, // syntect::html::ClassStyle::SpacePrefixed requires prefix to be &'static str
 }
 
 impl SyntectAdapter {
@@ -29,6 +30,7 @@ impl SyntectAdapter {
             theme: theme.map(String::from),
             syntax_set: SyntaxSet::load_defaults_newlines(),
             theme_set: ThemeSet::load_defaults(),
+            css_class_prefix: None,
         }
     }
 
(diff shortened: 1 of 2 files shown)


kivikakk:
Hiya! Thank you so much for this issue and fix!

I'm really happy with this, except for the `.clone().leak()`. I feel like this doesn't belong in library code.

Given the unavoidable `prefix: &'static str` from Syntect, could we instead please take an `Option<&'static str>` as the Comrak option? Then library consumers can choose to `.leak()` their input if that's what must be done, and can control when/how often it happens. As-is, it'll get called once _per highlighted code block_ every render, which could easily (but subtly) cause quite a lot of leaked memory over a server process's lifetime!

SteveBinary:
Thanks for your really quick response!
You are right, from a libraries point of view it makes totally sense to let the user decide on the leak.
I'll change that.

And as it turns out, `String::leak()` isn't even supported by your MSRV.

kivikakk:
Awesome. Thank you so much!

Issue #592, opened 2025-09-11
Superscript and subscript error when string begins with [^0-9a-z]


kivikakk:
From @ahamelers at gjtorikian/commonmarker#390:

> With the superscript and subscript extensions turned on (`extension: { superscript: true, subscript: true }`), any string that begins with a character that is not a letter or number is not correctly superscripted or subscripted. Instead, the characters `~~` or `^^` are escaped.
>
> For example (`m^-3^`), should be m<sup>-3</sup>. Instead it is rendered as m^-3^.
>
> I don't know if this problem is inherited—I'm happy to make an issue on https://github.com/kivikakk/comrak if so

```console
$ echo 'm^-3^' | cargo run -- -e superscript -e subscript
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/comrak -e superscript -e subscript`
<p>m^-3^</p>

$ echo 'm^3^' | cargo run -- -e superscript -e subscript
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.03s
     Running `target/debug/comrak -e superscript -e subscript`
<p>m<sup>3</sup></p>
```

kivikakk:
Fix was released in 0.43.0.

PR #593, opened 2025-09-11
superscript: failing test for #592.

kivikakk:
A successful fix for #592 should resolve this test.


kivikakk:
This occurs for the same reason `m*-a*` doesn't render an emphasis, but `m*a*` does; see ["can open emphasis"](https://spec.commonmark.org/0.31.2/#can-open-emphasis) and ["left-flanking delimiter run"](https://spec.commonmark.org/0.31.2/#left-flanking-delimiter-run) in the spec. The delimiter in `m*-` (or `m^-`) are is considered left-flanking because the delimiter (`*` or `^`) is followed by a Unicode punctuation character (`-`) but not preceded by a Unicode whitespace or Unicode punctuation character (`m`).

Unclear what a good solution is; we can redefine left-flanking for this case only, but it feels like quite a hack and perhaps it'd be better to just use actual math support in these cases, if inline HTML isn't desirable.

kivikakk:
I've downgraded my opinion of the hackiness of my proposed solution here; I think permitting punctuation following a delimiter, in the case where superscript/subscript are enabled and the delimiter in question is being scanned, is quite OK!

PR #594, opened 2025-09-12
Cleanup fuzzers, add unresolved `relaxed_autolink_email_in_footnote` test

Mrmaxmeier:
Hi,

I noticed that `comrak`'s fuzzing targets are somewhat outdated and fail to build.  The fixes to the harnesses are quite small (df59cf3f134050ede707aad0b4ad833446c81219) but I've also included some cleanup commits and two new harnesses.

While experimenting with fuzzers I ran into the following bug: I've included it as a `#[should_panic]` test since I'm not sure what the intended behaviour is. :upside_down_face:

```rust
[test]
// FIXME
#[should_panic = "assertion failed: (sp.end.column - sp.start.column + 1 == x) || rem == 0"]
fn relaxed_autolink_email_in_footnote() {
    assert_ast_match!(
        [
            extension.autolink,
            extension.footnotes,
            parse.relaxed_autolinks
        ],
        "[^a@b.c\nA]:\n",
        (document (1:1-1:1234) [
            // TODO: what should this be parsed as?
        ]),
    );
}
```

Let me know if I should drop any of the commits -- clippy's suggestions seem reasonable though I can understand if somebody doesn't like them.

### DIFF
--- a/src/parser/table.rs
+++ b/src/parser/table.rs
@@ -96,8 +96,7 @@ fn try_opening_header<'a>(
         let header_ast = &mut header.data.borrow_mut();
         header_ast.sourcepos.start.line = start.line;
         header_ast.sourcepos.end = start.column_add(
-            (container.data.borrow().content.len() - 2 - header_row.paragraph_offset)
-                as isize,
+            (container.data.borrow().content.len() - 2 - header_row.paragraph_offset) as isize,
         );
     }
 


kivikakk:
Hi there! Thank you for this! It's true that I mostly just test on the `all_options` fuzzer these days; I'm very happy to have these all back in order.

I've looked through the clippy cleanups and I'm happy to accept them all. IIRC in the past there were some suggestions I used to leave un-applied because the changes weren't compatible with our MSRV, but I can build and test everything in this branch successfully on 1.65 so LGTM :)

I'll think on the FIXME you've added and figure out what could reasonably be intended! @digitalmoksha might also have an opinion as the author of `relaxed_autolinks`.

Thanks again, this is a lovely PR 💜

PR #607, opened 2025-09-24
Add CJK Friendly Emphasis to CLI option

tats-u:
I have forgot to add this as a CLI option.


kivikakk:
Good catch, thank you!

PR #608, opened 2025-09-24
Introduce `cargo binstall`

tats-u:
[cargo binstall](https://github.com/cargo-bins/cargo-binstall) makes it possible to install CLIs in a crate without compilation. It just find and download an installer or archive in the release page in the repository of the crate instead. If it is not possible, it falls back to the normal compilation.


kivikakk:
Neat — thank you!

PR #616, opened 2025-10-05

kivikakk:
rust-analyzer reports only 664 errors to go!

imagen

The big issue will be reworking the main block and inline parsers to not layer mutable borrows — but if I'm successful, it means we won't have to deal with dynamic borrows at all! The codebase would become _much_ clearer, not to mention faster.

To be continued when I next get time!


kivikakk:
Down to 33.

kivikakk:
This was not, in fact, merged. See #617.

PR #618, opened 2025-10-09
Make validation of autolinks stricter

chamlis:
This PR aims to address two cases where autolink is too eager, by both my intuition and reading of the spec. In practice, they've come up [on Lobsters](https://github.com/lobsters/lobsters/issues/1727), where [commonmarker](https://github.com/gjtorikian/commonmarker) is used to render Markdown across the site.

The first commit covers an issue with www autolinks. Per the spec, "www." should be followed by a valid domain, but the domain validation includes the initial "www.". Valid domains must include at least one period, but this condition would always be fulfilled by the initial "www.". This means that both "www." (an empty domain) and "www.com" (a domain without a period) were autolinked.

This second commit covers an issue where the number of periods in a domain was not validated for an extended URL autolink. This was clearly intentional, as there was a dedicated parameter to loosen the check, and tests for it. However, it means that bare schemes such as "http://" are autolinked. I can't see any reference to this modified validation in the spec.

I've opened this as a draft because these changes, particularly the second, affected a lot of tests, which I've patched up for now. I don't want to break anybody who depends on that behaviour. Maybe some different behaviour would be ideal, or maybe I've misread the spec entirely!

### DIFF
--- a/src/parser/autolink.rs
+++ b/src/parser/autolink.rs
@@ -409,7 +409,7 @@ pub fn url_match<'a>(
         }
     }
 
-    let mut link_end = check_domain(&contents[i + 3..], relaxed_autolinks)?;
+    let mut link_end = check_domain(&contents[i + 3..], relaxed_autolinks)? + 3;
 
     while link_end < size - i && !isspace(contents[i + link_end]) {
         // basic test to detect whether we're in a normal markdown link - not exhaustive


kivikakk:
Thanks for this PR! It'll take me a little while to figure out what exactly to do here.

To give you some context, the GFM extensions to the CommonMark spec were originally authored by me; they haven't been updated since I left the company, and as you've seen, `cmark-gfm` also looks mostly abandoned at this point[^github]. As a result, there's certainly scope for "interpreting" the spec beyond the executable examples in the primary spec text itself.

There's also a supplementary executable spec from upstream now failing: https://github.com/kivikakk/comrak/actions/runs/18385875475/job/52498634166?pr=618#step:5:256. It's asserting `http://inlines` should be autolinked, which you've correctly observed is contrary to the spec text. Again, I wrote that executable spec, the definition of "valid domain" from the spec text, and the original code in cmark-gfm conforming to it, and so that same behaviour followed into Comrak since we check we pass all of "upstream"'s tests. (And sure enough, GitHub does link http://inlines.)

I don't assume I'd have any luck trying to get that file changed, so there's a bit of a question of "what to do when the spec is named after and controlled by a company[^company] uninterested in maintaining it?" Particularly when it contradicts itself in such a manner! There's a bigger existential question unearthed here.

I think I basically agree with half of the changes, though I think that `allow_short` parameter was doing something necessary — I know there are users depending on that `smb:///` example, though that should be possible only under `relaxed_autolinks`, so we should be able to find a way to keep that spec as-is without affecting the default autolinks behaviour. Likewise the IPv6 addresses — that can get moved into `relaxed_autolinks`, preserving default autolinks' spec-compliancy.

The `*fuzz*` specs should remain unchanged in their input — they were created by fuzzers usually represent inputs that caused the program to crash. If they no longer actually result in autolinks, that's fine, we just need to adjust their outputs.

Similarly/oppositely, the `autolink_sourcepos` test input should be adjusted to construct autolinks (since we're asserting not whether or not autolinks are created, but what the sourcepos of actually autolinked links are!), so `www.com` should be changed to something that autolinks instead.

I'm not asking you to make any of these changes — just documenting my thoughts for when I get back to it, though I welcome your input if you have further thoughts!

Thanks for opening this. Once we get it straightened out a bit, I can check how GitLab's full CI suite goes with the changes; Comrak is used for all Markdown processing there, the suite is really big, and should unearth anything I wasn't able to unanticipate! It might be there needs to be more granularity than just "autolinks" and "relaxed autolinks" to support downstream use cases without making the default autolinks non-compliant with the spec — but again, I'm really scratching my head about the value of an underspecified spec, without maintainers, with an supplementary executable spec that contradicts the primary spec's text.

[^github]: That said, I don't _believe_ they use Comrak (yet?) for their backend — hence the differences in autolinking behaviour between GitHub and Comrak. They do do extensive post-processing, though, so who knows.

[^company]: Or not even a company; an almost-division of a company.

chamlis:
Thanks for such a comprehensive reply!

My original plan was to try and get a similar change into `cmark-gfm` before `comrak`, to reduce how much the latter deviates in behaviour. As you say, `cmark-gfm` hasn't had updates in a while, so I decided to try more directly. I don't suppose you know of any canonical repository for the spec itself, where this kind of discussion might be better suited?

Apologies on the failing test, I was only running `cargo test`. I'm happy to make those suggested fixes wrt. the fuzzing and sourcepos tests if and when we know what the behaviour should be.

In my newcomer's opinion, I see three approaches:
- Move the current behaviour behind `relaxed_autolinks` as you've suggested, and try to strictly implement the spec as written by default.
- Patch the spec (if only downstream) to explicitly support IPv6 and SMB URLs. Strictly speaking I guess the latter should only happen under `relaxed_autolinks` anyway as it's not one of the schemes mentioned in the spec, so that would be covered by the first approach. I don't know what menagerie of other URLs users expect to autolink, so this might not be feasible in general.
- Only make things slightly stricter by explicitly forbidding valid domains to be zero-length. This would use the first commit here, and do an extra check when we have `allow_short` and `np=0`. This would fix the Lobsters issues, and should minimise any downstream breakage (until such time as `www` is a TLD, I don't think this would stop linking any valid URLs).

Thanks again; I'm happy to chime in with suggestions and do some implementation, but I appreciate that you're the primary source on much of this and am also happy to defer those decisions accordingly.

kivikakk:
Hey! Thank you for the quick reply and lovely discussion.

> I don't suppose you know of any canonical repository for the spec itself, where this kind of discussion might be better suited?

AFAICR, the spec website is deployed from a private repository (possibly `github/gfm` to correspond with <https://github.github.com/gfm>), with `cmark-gfm`'s `test/spec.txt` copied in and a `Makefile` run to generate the HTML. That HTML is committed to a branch which GitHub Pages deploys.

I don't know of any public place to discuss, and given the lack of activity on `cmark-gfm`, I kind of suspect there's maybe no-one at the company these days who actually is responsible for maintaining it any more? The `cmark-gfm` repository was the place to have any such discussions when I was working on it. The last time I asked for a spec update since leaving was there, too, 3 years ago; see https://github.com/github/cmark-gfm/issues/271.

(That said, I (personally) won't be doing that again, as (a) I don't really want to keep myself beholden to the whims of a company not willing to invest in their "property" in this way, and (b) I don't want to keep leaning on my connections and bugging my lovely former coworkers who aren't paid or recognised for doing such extra work.)

> Apologies on the failing test, I was only running `cargo test`. I'm happy to make those suggested fixes wrt. the fuzzing and sourcepos tests if and when we know what the behaviour should be.

Not at all a problem, CI didn't run until I hit the approve button so I wasn't expecting you to find it. :) 

Looking at the three suggestions:

> Move the current behaviour behind `relaxed_autolinks` as you've suggested, and try to strictly implement the spec as written by default.

The bigger story with `relaxed_autolinks` is that it was implemented (afaict) essentially to make things easier for one downstream consumer, GitLab. (See https://github.com/kivikakk/comrak/pull/325, which first introduced it just to recognise autolinks in brackets, and then https://github.com/kivikakk/comrak/pull/380, which added the "any scheme" part.) That said, it seems to have many more users now — I see a few downstream libraries suggesting it as a default option — so we should keep supporting as permissive a parsing as makes sense under that option, at the very least.

(If changes mean `relaxed_autolinks` doesn't parse `http://`, that's 100% OK, since I don't think that makes sense, but it should continue to parse whatever scheme and domains without periods.)

> Patch the spec (if only downstream) to explicitly support IPv6 and SMB URLs. Strictly speaking I guess the latter should only happen under `relaxed_autolinks` anyway as it's not one of the schemes mentioned in the spec, so that would be covered by the first approach. I don't know what menagerie of other URLs users expect to autolink, so this might not be feasible in general.

As for _why_ GitLab want to parse any scheme, since 4 weeks ago I now have the insider view (!) — it's because companies deploy GitLab in their intranet and want to link whatever strange and wonderful internal things they like. SMB and IPv6 are part of it but we can essentially never presume to know the full list, so I think we can't be restrictive, and patching the spec downstream would be end up being Sisyphean on multiple fronts.

> Only make things slightly stricter by explicitly forbidding valid domains to be zero-length. This would use the first commit here, and do an extra check when we have `allow_short` and `np=0`. This would fix the Lobsters issues, and should minimise any downstream breakage (until such time as `www` is a TLD, I don't think this would stop linking any valid URLs).

I'm totally fine with this as the quickest route to a fix.

tl;dr: I think the first or last option are great and I would happily accept a PR that did either of these.

chamlis:
I've opted for the first approach, and force-pushed and marked ready for review as appropriate.

I tried the third approach, but for IPv6 autolinks the valid domain is also zero-length - it's just all included as part of the link's path. We'd then have to allow zero-length domains in relaxed mode, but we get the same effect through the first approach by requiring at least one period in strict mode.

It's certainly a neater patch this time around. I hope I've preserved the intent of the tests properly, otherwise I'm happy to amend and force-push again.

A quick note: I think the [other call to `valid_domain`](https://github.com/kivikakk/comrak/blob/6d22e82f80336cbeed06e07110fd6316fc655247/src/parser/autolink.rs#L411) should have a similar `+ 3` on the end, to account for the `://` we're skipping for validation but including in the link. I don't think this will ever cause any issues, as the link extension just below is more lenient so will include the missing three characters anyway.

Thanks once more for all of your insight and feedback on this.

kivikakk:
Thank you! I've made a slight adjustment by changing the autolink sourcepos test to not rely on relaxed autolinks; it's measuring the baseline autolink sourcepos behaviour, so it's better for it to not rely on anything else.

I've also adjusted our CI script so we use `--relaxed-autolinks` when testing autolink behaviour; this gets us past the questionable/contradictory executable spec.

> A quick note: I think the [other call to `valid_domain`](https://github.com/kivikakk/comrak/blob/6d22e82f80336cbeed06e07110fd6316fc655247/src/parser/autolink.rs#L411) should have a similar `+ 3` on the end, to account for the `://` we're skipping for validation but including in the link. I don't think this will ever cause any issues, as the link extension just below is more lenient so will include the missing three characters anyway.

Yep, there's no reason it shouldn't skip them; thanks!

Thanks so much for your contributions, both in discussion and code! I really appreciate it!

kivikakk:
I'll put out a release shortly. :)

kivikakk:
This is part of [Comrak 0.44.0](https://crates.io/crates/comrak/0.44.0). I'll make a Commonmarker release later today too!

kivikakk:
[Commonmarker 2.5.0](https://rubygems.org/gems/commonmarker/versions/2.5.0) has been released!

```ruby
> Commonmarker::VERSION
=> "2.5.0"
> Commonmarker.to_html('hi www.com http:// yay', options: { extensions: { autolink: true } })
=> "<p>hi www.com http:// yay</p>\n"
> Commonmarker.to_html('hi www.com http:// yay', options: { extensions: { autolink: true }, parse: { relaxed_autolinks: true } })
=> "<p>hi <a href=\"http://www.com\">www.com</a> <a href=\"http://\">http://</a> yay</p>\n"
```

Thanks again for your help! 🤍

PR #624, opened 2025-10-18
syntect optional in cli; disable syntax highlighting in bench.

kivikakk:
Also remove process creation overhead, and make it easier to use samply for profiling.


kivikakk:
`run_benchmarks` fails because this PR tries to build `comrak-main` without Syntect, which it won't support until this PR is merged, but the difference is significant:

image

This comparison is finally a little more even-handed.

PR #627, opened 2025-10-18
Refactor, simplify, speed up.

kivikakk:
This mostly just applies to smart quotes, which are unlikely to benefit hugely (small String optimisation?), but I'm interested in seeing how far we can go with Cow; what if we can avoid allocations entirely in some cases by referring to the input?


kivikakk:
Rewritten an incredible amount of 2018-era C-isms and laid bare the control flow in a bunch of places! Soooo much nicer.

PR #628, opened 2025-10-18

kivikakk:
Keeping the dream alive, by which I mean keeping this rebased in case there's something still useful here. I'll profile it some more in a bit.

See also #616, #617.


kivikakk:
At the moment it's still a slowdown:

```
Benchmark 1: ./bench.sh ./comrak-d4fb6782
  Time (mean ± σ):      84.2 ms ±   1.7 ms    [User: 65.1 ms, System: 19.8 ms]
  Range (min … max):    80.7 ms …  88.5 ms    33 runs

Benchmark 2: ./bench.sh ./comrak-main
  Time (mean ± σ):      75.4 ms ±   0.5 ms    [User: 59.4 ms, System: 16.9 ms]
  Range (min … max):    74.1 ms …  76.7 ms    38 runs

Summary
  ./bench.sh ./comrak-main ran
    1.12 ± 0.02 times faster than ./bench.sh ./comrak-d4fb6782
```

PR #634, opened 2025-10-19
try aarch64 simd jetscii with const new.

kivikakk:


kivikakk:
This is neat, but belongs upstream.

PR #635, opened 2025-10-19
test: add complex tilde behavior

miketheman:
According to the `cmark-gfm` spec, using 3 or more tildes should not become a strikethrough.
Refs: https://github.github.com/gfm/#example-493

However, using this kind of string on GitHub.com today adds the `del` attributes.

This test will fail, but `cmark-gfm` behaves differently.
Which one is correct?


kivikakk:
Correct according to what, I suppose? 😅 As you've noticed, the spec text isn't very clear on the edge cases (which is my fault!! sorry!).

The last relevant strikethrough change on cmark-gfm (producing the results you see on GitHub) are here: https://github.com/github/cmark-gfm/pull/120/files

I'd be inclined to bring Comrak into line, given that and the discussion in https://github.com/github/cmark-gfm/issues/118.

Are you interested in contributing the fix as well? If so, I'll leave it to you. If not, I'm happy to pick it up.

miketheman:
@kivikakk Thanks for the pointers, I've added a commit with what I believe is a fix. Not super familiar with Rust conventions, so let me know if it needs any updates.

kivikakk:
Thank you so much! LGTM :)

miketheman:
Excellent!
Will this make it into the upcoming 0.45 release?

kivikakk:
Yes, it will!

PR #646, opened 2025-10-23
fix: Incorrect `sourcepos` for escaped character spans

### LINKED ISSUE #562: incorrect `sourcepos` for escaped character spans
_First off, shoutout to @kivikakk for the recent inline sourcepos fixes! I ran into a bug with source positions recently and just thought - **oh, no** - until I updated from 0.29 to 0.38 and not only was my issue resolved but I also got subscripts and alerts. Nice!_

This is another one of #448, #502, #507. When using `options.render.escaped_char_spans = true;`, the sourcepos info for the text child node in the escaped tag node contains the backslash that escaped it. This is in contrast to, say, the text child node of an emph node, where the emph includes the syntax that made it emph but the text inside doesn't.

So, in this example doc...
```md
\`code\`
```
...my program's ascii-art debug output is...
```
■ Document 1:1-1:8 +1■
  └>■ Paragraph 1:1-1:8 +3=
    ├>= Escaped 1:1-1:2 +1■
    │ └>= Text("`") 1:1-1:2
    ├>= Text("code") 1:3-1:6
    └>= Escaped 1:7-1:8 +1■
      └>= Text("`") 1:7-1:8
```
...but I expect it to be...
```
■ Document 1:1-1:8 +1■
  └>■ Paragraph 1:1-1:8 +3=
    ├>= Escaped 1:1-1:2 +1■
    │ └>= Text("`") 1:2-1:2
    ├>= Text("code") 1:3-1:6
    └>= Escaped 1:7-1:8 +1■
      └>= Text("`") 1:8-1:8
```
...with the start and end LineColumn's of each text range being equal, consistent with them having length 1, and with those text ranges being strict subsets of the text ranges for their parent 'escaped' nodes.

Martin005:
This PR fixes the incorrect source position tracking for escaped characters and updates related tests to reflect these changes.

Fixes #562

### DIFF
--- a/src/parser/inlines.rs
+++ b/src/parser/inlines.rs
@@ -392,7 +392,7 @@ impl<'a, 'r, 'o, 'd, 'c, 'p> Subject<'a, 'r, 'o, 'd, 'c, 'p> {
 
             let inline_text = self.make_inline(
                 NodeValue::Text(self.input[self.pos - 1..self.pos].to_string().into()),
-                self.pos - 2,
+                self.pos - 1,
                 self.pos - 1,
             );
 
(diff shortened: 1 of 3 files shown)


kivikakk:
Wonderful! Thank you so much, especially for the quality tests added.

PR #648, opened 2025-10-24
Add support for Discord style subtext

Kuuuube:
Fixes #647

Unsure if this needs cleanup. Tried to follow the existing codebase and style as closely as possible.

Also noticed that the project was failing to build (before I made any changes) without a `cargo update`. I didn't push the cargo lock changes but I can if you'd like.

Not using "Discord" in the docs since I noticed #421 did not mention Discord by name anywhere. Unsure if this is an intentional choice and there's some reason for omitting it or not.

PR #650, opened 2025-10-24
Fix subtext having subscript option in doc example and add cfg_attr

Kuuuube:
Tiny thing I messed up in #648

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -803,7 +803,11 @@ where
     }
 
     fn detect_atx_subtext(&self, line: &str) -> Option<usize> {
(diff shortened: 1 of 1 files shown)


kivikakk:
Thanks for the fix! Though this makes me think; how did the doc-tests pass if we weren't actually enabling the option? We forgot to gate the parsing functionality on the option! Could you add a condition in `Parser::detect_atx_subtext` while we're here?

Kuuuube:
Oops. That's what I get for looking off of the header code for that. Sorry for the trouble. Tests now appear to correctly fail when `subtext` is not set.

kivikakk:
:D Tysm!

PR #654, opened 2025-10-26
fix: unicode isolation should terminate links

SethFalco:
I'm using Project Fluent for localization in some of my projects. The output is plaintext, and then the client parses it (often as Markdown with autolinking).

Project Fluent wraps all interpolated variables with direction isolation by default, which is a great feature I'd like to keep enabled. This reduces the risk of changes in language direction from messing with presentation of text.

> Without isolation, the rendered result shows the trailing slash on the wrong side of the URL:
>
> > تفضل بزيارة https://www.mozilla.org/privacy/ لمعرفة المزيد. ❌
>
> With isolation, the forward slash is displayed correctly:
>
> > تفضل بزيارة ⁨https://www.mozilla.org/privacy/⁩ لمعرفة المزيد. ✅
> 
> \- [Unicode Isolation: What does it look like?](https://github.com/projectfluent/fluent.js/wiki/Unicode-Isolation#what-does-it-look-like)

However, some Markdown parsers such as Comrak do not treat Unicode isolation as terminating characters for URLs, so URLs break for clients. Try visiting the URL quoted above with the ✅! :fearful:

This solves it by treating `\u{2069}` as a terminating character. I originally implemented this as an array of Unicode characters, but as performance is a big consideration for this library, and it's unclear if additional non-ASCII characters would ever be required, I amended this to hard code the one character.

It should be easy to migrate the `[u8; 3]` to `[[u8; 3]; 2]` or whatever whenever the need arises.

## Before and After

```diff
Input:
\u{2068}https://www.example.com/\u{2069}

Output:
- <p>\u{2068}<a href=\"https://www.example.com/%E2%81%A9\">https://www.example.com/\u{2069}</a></p>\n
+ <p>\u{2068}<a href=\"https://www.example.com/\">https://www.example.com/</a>\u{2069}</p>\n 
```

## ⚠️ Notes

I only have superficial knowledge of Rust. Mentorship would be very appreciated if you have any gripes with my implementation. :pray:

## Related

* https://github.com/projectfluent/fluent/discussions/375

### DIFF
--- a/src/parser/autolink.rs
+++ b/src/parser/autolink.rs
@@ -376,12 +376,14 @@ fn autolink_delim(data: &str, mut link_end: usize, relaxed_autolinks: bool) -> u
 
             link_end -= 1;
         } else if cclose == LINK_END_UNICODE[2] {
-            let slice = &bytes[link_end - LINK_END_UNICODE.len() .. link_end];
+            let slice = &bytes[link_end - LINK_END_UNICODE.len()..link_end];
 
             if slice == LINK_END_UNICODE {
                 link_end -= LINK_END_UNICODE.len();
                 break;
             }
+
+            break;
         } else {
             break;
         }


src/parser/autolink.rs
```diff
@@ -373,6 +375,13 @@ fn autolink_delim(data: &str, mut link_end: usize, relaxed_autolinks: bool) -> u
             }
 
             link_end -= 1;
+        } else if cclose == LINK_END_UNICODE[2] {
+            let slice = &bytes[link_end - LINK_END_UNICODE.len() .. link_end];
+
+            if slice == LINK_END_UNICODE {
```
kivikakk: If I were writing this today, I'd probably use something like:

```rust
if bytes[..link_end].ends_with(LINK_END_UNICODE) {
```

This avoids a problem if the link ends in 0xa9, but there aren't actually two more bytes before it. This isn't actually possible thanks to the combination of circumstances that leads us here: the shortest UTF-8 sequence ending in 0xa9 is two bytes (e.g. 0xc2 0xa9 for U+00A9 COPYRIGHT SIGN), and that alone will never be enough to trigger `autolink_delim`, so there'll always be a third byte.

Still, I mention it because missing length checks in places like this often _do_ end up being the cause of a surprise unsigned underflow panic, and it's nice to write in a way that avoids it where possible.

(I also have to apologise for much of the code in this module; it's not written very well, so I wouldn't learn too much from it if I were you!)

src/parser/autolink.rs
```diff
@@ -373,6 +375,13 @@ fn autolink_delim(data: &str, mut link_end: usize, relaxed_autolinks: bool) -> u
...
+                link_end -= LINK_END_UNICODE.len();
+                break;
+            }
```
kivikakk:
```suggestion
            }
            break;
```

Without the `break;` (per the catch-all `else`), any encounter of a trailing 0xa9 byte that isn't part of U+2069 POP DIRECTIONAL ISOLATE will cause an infinite loop. e.g.:

```console
$ printf 'www.copyrighted.com\xc2\xa9' | cargo run -- -e autolink
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.07s
     Running `target/debug/comrak -e autolink`
```

kivikakk:
> I originally implemented this as an array of Unicode characters, but as performance is a big consideration for this library, and it's unclear if additional non-ASCII characters would ever be required, I amended this to hard code the one character.

Indeed :) I like the "minimum viable change" aspect of this PR. :) Thank you!

Issue #655, opened 2025-10-27
Parsing for HTML block of kind 4 expects only ASCII uppercase letters


Martin005:
When implementing the tests in #656, I noticed the following issue.

The [CommonMark spec about HTML block](https://spec.commonmark.org/0.31.2/#html-blocks) of kind 4 says:

> 4. **Start condition**: line begins with the string `<!` followed by an ASCII letter.
> **End condition**: line contains the character `>`.

When looking at the following input, both lines should create a HTML block according to the spec:
```
<!A>
<!a>
```

I tried it in the [commonmark.js dingus](https://spec.commonmark.org/dingus/?text=%3C!A%3E%0A%3C!a%3E) and the AST output was:
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE document SYSTEM "CommonMark.dtd">
<document xmlns="http://commonmark.org/xml/1.0">
  <html_block>&lt;!A&gt;</html_block>
  <html_block>&lt;!a&gt;</html_block>
</document>
```

But when I tried it in the [CommonMark/GFM/GLFM dingus](https://gitlab-org.gitlab.io/ruby/gems/gitlab-glfm-markdown/?text=%3C!A%3E%0A%3C!a%3E#result-ast) (the only dingus online that uses comrak - even though it uses v0.43, I tested this locally with the most up-to-date `main` branch and it behaved the same), the AST output was:
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE document SYSTEM "CommonMark.dtd">
<document sourcepos="1:1-2:4" xmlns="http://commonmark.org/xml/1.0">
  <html_block sourcepos="1:1-0:0" xml:space="preserve">&lt;!A&gt;
</html_block>
  <paragraph sourcepos="2:1-2:4">
    <text sourcepos="2:1-2:4" xml:space="preserve">&lt;!a&gt;</text>
  </paragraph>
</document>
```

I think the culprit is the [line 180 in scanners.rs](https://github.com/kivikakk/comrak/blob/ad4bc940d3b6452759a3be958cf64a9831d74e7b/src/scanners.re#L180) that expects the ASCII letter to be uppercase:
```
'<!' [A-Z] { return Some(4); }
```

kivikakk:
Nice catch! This error has existed since [the very first version of Comrak](https://github.com/kivikakk/comrak/commit/9b6c5bf64cdaf5b4d91d02eb4dd38bfd885f39f9):

[...]

PR #656, opened 2025-10-27
fix: Incorrect `sourcepos` for HTML blocks

### LINKED ISSUE #448: `sourcepos` is incorrect for `<script>` tags
I've encountered an odd issue. I'm using the Ruby wrapper (Commonmarker) and inspecting the AST shows some strange behavior around the `sourcepos` of `<script>` tags. AFAICT this is only related to script tags, and only ones that appear all on one line at that.

```ruby
Commonmarker.parse("<script></script>")
=> #<Commonmarker::Node(document):
  source_position={:start_line=>1, :start_column=>1, :end_line=>1, :end_column=>17}
  children=[#<Commonmarker::Node(html_block):
       source_position={:start_line=>1,
        :start_column=>1,
        :end_line=>0,
        :end_column=>0}>]>
```
note the `end_line` and `end_column` are both zero. In general I've observed that `end_line` is `start_line - 1`, and `end_column` is always zero.

Let me know if I should open an issue on Commonmarker instead, but I don't see how this could be caused by the wrapper. I just lack the rust expertise to test it directly in rust, sorry!

### LINKED ISSUE #507: Incorrect `sourcepos` for `HtmlBlock` of a html comment
source:

* expected: `1:1-1:21`
* actual: `1:1-0:0`

```rust
let text = "<!-- A html block -->";
let arena = Arena::new();
let ast = parse_document(&arena, text, &Options::default());
println!("{:?}", ast);
```

Martin005:
This PR fixes incorrect `sourcepos` for HTML blocks.

When writing the tests, I came across a bug, so I created this issue as a follow-up: #655 

Fixes #448, fixes #507

### DIFF
--- a/src/parser/inlines.rs
+++ b/src/parser/inlines.rs
@@ -20,7 +20,7 @@ use crate::parser::options::{BrokenLinkReference, WikiLinksMode};
 use crate::parser::shortcodes::NodeShortCode;
 use crate::parser::{autolink, AutolinkType, Options, ResolvedReference};
 use crate::scanners;
-use crate::strings::{self, is_blank, Case};
+use crate::strings::{self, count_newlines, is_blank, Case};
 
 const MAXBACKTICKS: usize = 80;
 const MAX_LINK_LABEL_LENGTH: usize = 1000;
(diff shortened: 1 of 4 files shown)


kivikakk:
Thank you — amazing. I've refactored the newline counter with another similar function and put it somewhere hopefully more discoverable.

Martin005:
@kivikakk Thank you! That refactoring makes sense, thanks again ❤️

PR #679, opened 2025-11-02
prevent unexpected post-processing & simplify internal feed.

kivikakk:
Some minor fixes and cleanups, and the main course: prevent unexpected post-processing when elements are escaped.

This brought to my attention by `- \[x] hello` with tasklist (only) enabled raising an assertion: it's still trying to parse it as a tasklist, and the tasklist code asserts the text element's sourcepos start matches the containing paragraph's. In this case, it doesn't: the escape isn't included, so it throws for a mismatch.

The solution here _seems_ to be to let the `Escaped` node actually always enter the document tree, preventing this from getting any post-processing treatment at all. We remove them after Text node unification and post-processing (unless the user wants them left in).

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -136,6 +136,11 @@ where
         self.total_size = end;
 
         let mut ix = 0;
+
+        // linebuf is necessarily entirely to do spec-compliant NUL handling in
+        // one place. If the input document contains no NUL bytes, we will never
+        // use linebuf. Our re2c scanners presume there are no NUL bytes in
+        // the subject, and use 0 as the sentinel result when !(cursor < len).
         let mut linebuf = String::new();
 
         while ix < end {
@@ -163,10 +168,13 @@ where
             if ate_line_end > 0 || eol == end {
                 if !linebuf.is_empty() {
                     linebuf.push_str(&s[ix..eol]);
-                    let line = mem::take(&mut linebuf);
-                    self.process_line(line.into());
+                    // Keep one active linebuf allocation.
+                    let mut cow = Cow::Owned(mem::take(&mut linebuf));
+                    self.process_line(&mut cow, eol == end);
+                    mem::swap(&mut cow.into_owned(), &mut linebuf);
+                    linebuf.clear();
                 } else {
-                    self.process_line(s[ix..eol].into());
+                    self.process_line(&mut s[ix..eol].into(), eol == end);
                 }
             } else {
                 assert_eq!(sb[eol], b'\0');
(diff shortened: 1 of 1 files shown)


kivikakk:
Cursed. The following Markdown:

```
a.b-c_d@a.b_
```

formats to CommonMark as follows:

```
a.b-c\_d@a.b\_
```

The status quo is that this is post-processed the same as the original input; one big text node with the escapes parsed out/removed:

* document
  * paragraph
    * text `a.b-c_d@a.b_`

With autolinks turned on, these aren't autolinked; the trailing underscore disqualifies it (I guess).

With this PR, the roundtripped CommonMark parses as:

* document
  * paragraph
    * text `a.b-c`
    * escaped
      * text `_`
    * text `d@a.b`
    * escaped
      * text `_`

This _does_ get parsed as an autolink, and so we fail a roundtrip test. Grr.

I'm kind of inclined to think this parse is actually correct (although different to the status quo), and it's the formatter that's changing the meaning here and should be corrected. But that's a Hard problem, as it's difficult to fully generally determine when a character needs to be escaped to be reinterpreted the same way, seeing as preceding and following characters will influence its interpretation.

kivikakk:
This change is necessary for `escaped_char_spans` to not bifurcate the interpretation of post-processed extensions in Comrak's eyes anyways: those `Escaped` elements being inserted only when the option is turned on visibly influence the interpretation. So we have to do it unilaterally, and remove the elements from the AST if _not_ wanted, if we don't want it to actually change the meaning of the document when turned on.

The alternative is to track such characters without modifying the tree (e.g. vec of character indices), but that becomes hard to maintain. Hrmmmm.

kivikakk:
> This _does_ get parsed as an autolink, and so we fail a roundtrip test. Grr.

I've decided this parse is correct: without the backslash, the `_` is part of the surrounding text and the [autolink rules](https://github.github.com/gfm/#example-631) say the underscore disqualifies it:

> `.`, `-`, and `_` can occur on both sides of the `@`, but only `.` may occur at the end of the email address, in which case it will not be considered part of the address:

With the backslash, (a) it's not a `_` at the end, it's a backslash, and (b) even if we say it's not a backslash, it's not a `_`, it's an _escaped_ `_`. The purpose of escaping is to prevent the usual interpretation.

kivikakk:
It occurs to me that the answer here is to _not_ keep the autolinks extension turned on while doing roundtrip tests: any GFM extension autolinks recognised in the original text will be written out as core spec links (of some variety). Doing further extension autolink parsing on the resulting roundtripped text may well not be stable, because we don't roundtrip GFM extension autolinks to begin with!

PR #682, opened 2025-11-03
don't append a newline at EOF; handle it in the parser.

kivikakk:
This gets rid of the last input feed oddity. The reference implementation forces every parsed line to end in a newline, even if the last line does not. This is not for spec reasons; [a line can be ended by the end of file](https://spec.commonmark.org/0.31.2/#line):

> A [line](https://spec.commonmark.org/0.31.2/#line) is a sequence of zero or more [characters](https://spec.commonmark.org/0.31.2/#character) other than line feed (U+000A) or carriage return (U+000D), followed by a [line ending](https://spec.commonmark.org/0.31.2/#line-ending) or by the end of file.

This PR adjusts the parser to be capable of dealing with end-of-input instead of assuming a newline will always be present. There's also a fair bit of "actually deal with CR or CRLF instead of assuming it's always LF" added.

### DIFF
--- a/src/arena_tree.rs
+++ b/src/arena_tree.rs
@@ -205,6 +205,13 @@ impl<'a, T> Node<'a, T> {
         self.last_child.set(Some(new_child));
     }
 
+    /// Append multiple new children to this node, after existing children.
+    pub fn extend(&'a self, new_children: impl IntoIterator<Item = &'a Node<'a, T>>) {
+        for child in new_children.into_iter() {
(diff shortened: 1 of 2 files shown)


kivikakk:
Extensively fuzzed, manually checked, tested against [`gitlab-glfm-markdown`](https://gitlab.com/gitlab-org/ruby/gems/gitlab-glfm-markdown), all OK.

PR #694, opened 2025-11-21
Add byte_matches helper to simplify get/map_or chaining tests in parser

liamwhite:
This is a really common pattern, used in a bunch of places, so I centralized the definition to try to help with verbosity. Not sure about the impact there.

Also flipped the greentext checks for better readability since they use the pattern (the function name was negated).


kivikakk:
Clean 😻  Thank you!!

PR #700, opened 2025-11-24
chore: apply clippy suggestions

### LINKED ISSUE #695: CI not failing on clippy warnings
I noticed some clippy warnings but CI does report success. Is that expected?

Image

Image

I can submit a fix if that's indeed a problem.

leandrocp:
Close #695

Hi @kivikakk here's a proposal to clean up clippy warnings. Most changes are "just" eliding lifetimes, I don't know if there's a preference but I'm happy to adjust if needed.

```
❯ rustc --version
rustc 1.70.0 (90c541806 2023-05-31)

❯ cargo --version
cargo 1.70.0 (ec8a8a0ca 2023-04-25)
```

If this is good and gets merded, should the CI fail on clippy warns/errors?

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -16,7 +16,7 @@ fn main() {
     // Sort it for binary search.
     let mut translated_entities = ENTITIES
         .iter()
-        .filter(|e| e.entity.starts_with("&") && e.entity.ends_with(";"))
+        .filter(|e| e.entity.starts_with('&') && e.entity.ends_with(';'))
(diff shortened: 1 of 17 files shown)


kivikakk:
This all looks good to me, thank you!

> If this is good and gets merded, should the CI fail on clippy warns/errors?

I think so. I think the intention, even, is that it should — [we run Clippy in CI](https://github.com/kivikakk/comrak/actions/runs/19634457963/job/56221707087#step:4:208), but I guess we don't actually fail on warnings!

PR #702, opened 2025-11-25
fix(heex): self closing tag on same line

leandrocp:
Hi @kivikakk here's a fix for a bug I found running the example in https://github.com/leandrocp/mdex/pull/274

Idk when the next version is coming out but I'll test more edge cases soon, either way please always feel free to tag me on any issue/PR related to Phoenix HEEx.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -112,6 +112,6 @@ jobs:
     - name: Obtain Rust
       run: rustup override set $MSRV
     - name: Check clippy
-      run: rustup component add clippy && cargo clippy
+      run: rustup component add clippy && cargo clippy -- -Dwarnings
     - name: Check formatting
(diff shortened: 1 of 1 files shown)


src/nodes.rs
```diff
@@ -6,7 +6,7 @@ use std::convert::TryFrom;
 
 use crate::arena_tree;
 #[cfg(feature = "phoenix_heex")]
-pub use crate::parser::phoenix_heex::NodeHeexBlock;
+pub use crate::parser::phoenix_heex::{HeexNode, NodeHeexBlock};
```
leandrocp: I had to export `HeexNode` to use on my NIF (to have Elixir/Rust integration)

kivikakk:
Nice! And yes, I will tag you if something comes up and I can't work out it :) MDEx would be most affected and I wouldn't wanna break it by accident!

PR #716, opened 2026-01-02
Optimize is_special_char by merging lookup tables

Martin005:
This PR optimizes `is_special_char` by merging lookup tables.

### Changes

- **Merged `smart_char_bytes` into `special_char_bytes`** during initialization when smart mode is enabled
- **Removed redundant `smart_char_bytes` field** from Subject struct
- **Eliminated 2 conditional branches** from the hot path, reducing the function to a single array lookup plus edge case

### DIFF
--- a/src/parser/inlines.rs
+++ b/src/parser/inlines.rs
@@ -48,7 +48,6 @@ pub struct Subject<'a: 'd, 'r, 'o, 'd, 'c, 'p> {
     no_link_openers: bool,
     special_char_bytes: [bool; 256],
     skip_char_bytes: [bool; 256],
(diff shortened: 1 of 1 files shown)


Martin005:
It actually does alter the performance: 285.8 ms -> 278.9 ms (~ 3% faster). You can check the [performance report](https://codspeed.io/kivikakk/comrak/branches/Martin005%3Adev) 😉

kivikakk:
Thank you very much! I will make a similar change in `main` and attribute you; please note in future the newly-added policy on LLM contributions in #718.

PR #717, opened 2026-01-02
Meagerly reduce heap allocations and improve hash function performance

gjtorikian:
I spent my holidays ideating on ways to improve the performance of comrak. Ultimately, I concluded that there would need to be a hard break of some of the API, and rather than having that discussion right now, I went in this direction: a smaller pile of cleanups.

Each commit here attempts to do one thing, and some new dependencies were introduced. Nothing results in zomg massive gains, but the cleanup is still nice to do imho. They should be self explanatory, but some highlights to pull include:

* using [phf](https://crates.io/crates/phf) to create a hash function for looking up entities during compilation
* using [rustc-hash](https://crates.io/crates/rustc-hash) which, apparently, is faster than the default provided by `std` because it's non-cryptographic (TIL; the crates page has more info)
* avoid heap allocations using `smallvec` for stack-based storage (the idea being that brackets with a nesting level less than 8—a presumably common cap—would benefit from this)
* various removals of `clone`s on paths that didn't need them

BTW, one additional minor change which would definitely breaking, and _would_ have large performance gains, is around the `SyntectAdapter`. Currently, it defaults to applying a theme, but I found that defaulting to applying TextMate classes performs better when processing triple backtick code blocks. Just a thought for a future break!


kivikakk:
> I spent my holidays ideating on ways to improve the performance of comrak. Ultimately, I concluded that there would need to be a hard break of some of the API, and rather than having that discussion right now, I went in this direction: a smaller pile of cleanups.

I spent my last "holidays" (between-jobs) period doing the same thing, and reaching the same outcome every time 😓 (#616, #617, #628, …). I actually fully pursued the hard breaks, though, but after spending honestly ridiculous amounts of time on it, it either ended up being _very_ very hard (and I didn't give up easily, as my reflog will attest), or proving >10% worse in overall performance 😓😓

Nonetheless, I'd welcome such changes in future if you happened to come by with some that did net an improvement (or even just a cleaner API without a perf loss, which is what I was eventually hoping for with those indextree attempts)!

Thanks so much for this! I'm very happy for cleanups and just Doing Things Better™. Also happy to change the defaults for SyntectAdapter, though I imagine that for those doing serious work with Comrak, they are probably not using it.

gjtorikian:
Yaaay I’m treading down well worn paths. 😆

kivikakk:
Haha, yep! Additionally, I also spent some long hours on multiple attempts trying to thread `Cow` through from the input side, seeing how much it'd be possible not to duplicate content from the input string. This also proved surprisingly and repeatedly capital-D Difficult. It did produce useful gains in the optimisations around not _always_ and immediately cloning and modifying the input line, however, which is a strange behaviour of the original `cmark`'s line feed — some of the most significant improvements came from this effort.

PR #719, opened 2026-01-15
fix(heex): components mixed with markdown

leandrocp:
Hi @kivikakk !

This fix a problem where the parser would stop processing markdown after phoenix components, for eg the code block below would not render correctly:

````
# Hello World

<.link>This is a link</.link>

```elixir
IO.puts("hello world")
```
````

My approach is to track ending tags so it knows what is heex and what's not.

### DIFF
--- a/src/tests/phoenix_heex.rs
+++ b/src/tests/phoenix_heex.rs
@@ -1028,7 +1028,7 @@ fn code_blocks_mixed_with_components() {
             "</.link>\n",
             "\n",
             "```rust\n",
-            "let result = ammonia::clean(\"<b><img src='' onerror=alert('hax')>I'm not trying to XSS you</b>\");\n",
+            "let result = ammonia::clean(\"<b><img src='' onerror=alert('attack')>I'm not trying to XSS you</b>\");\n",
             "```\n",
         ),
         concat!(
@@ -1039,7 +1039,7 @@ fn code_blocks_mixed_with_components() {
             "<.link navigate=\"/?sort=asc\" replace={false}>\n",
             "  Sort By Price\n",
             "</.link>\n",
-            "<pre><code class=\"language-rust\">let result = ammonia::clean(&quot;&lt;b&gt;&lt;img src='' onerror=alert('hax')&gt;I'm not trying to XSS you&lt;/b&gt;&quot;);\n",
+            "<pre><code class=\"language-rust\">let result = ammonia::clean(&quot;&lt;b&gt;&lt;img src='' onerror=alert('attack')&gt;I'm not trying to XSS you&lt;/b&gt;&quot;);\n",
             "</code></pre>\n",
         ),
     );

### CODE REVIEW

**Comment by kivikakk** [comment]
Nice job @leandrocp! One thing I noticed is that we're not exactly tracking opens and closes on a per-tag basis. For instance, given this input:

```markdown
<.abc> </.abc> <.xyz>

# Markdown!

</.xyz>
```

We'll process it as:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE document SYSTEM "CommonMark.dtd">
<document xmlns="http://commonmark.org/xml/1.0">
  <paragraph>
    <heex_inline xml:space="preserve">&lt;.abc&gt;</heex_inline>
    <text xml:space="preserve"> </text>
    <heex_inline xml:space="preserve">&lt;/.abc&gt;</heex_inline>
    <text xml:space="preserve"> </text>
    <heex_inline xml:space="preserve">&lt;.xyz&gt;</heex_inline>
  </paragraph>
  <heading level="1">
    <text xml:space="preserve">Markdown!</text>
  </heading>
  <paragraph>
    <heex_inline xml:space="preserve">&lt;/.xyz&gt;</heex_inline>
  </paragraph>
</document>
```

Or for example:

```markdown
<.abc>

</.abc><.xyz>

# Markdown!

</.xyz>
```

Parses as:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE document SYSTEM "CommonMark.dtd">
<document xmlns="http://commonmark.org/xml/1.0">
  <heex_block xml:space="preserve">&lt;.abc&gt;

&lt;/.abc&gt;&lt;.xyz&gt;
</heex_block>
  <heading level="1">
    <text xml:space="preserve">Markdown!</text>
  </heading>
  <paragraph>
    <heex_inline xml:space="preserve">&lt;/.xyz&gt;</heex_inline>
  </paragraph>
</document>
```

I'm happy to merge this now as it's an improvement on the baseline, but I figured I'd let you know about these cases I spotted! 😸

leandrocp:
Thanks for the quick review. Yep you're right, we're not tracking open/end exactly but as long as the heex tags are preserved then it should be fine, for eg  nested components like <.foo><.bar>... where <.bar> is detected as "text" (content) of the parent <.foo> component. The LiveView parser will handle it all as a piece of heex template. By tracking ending tags I meant the parser must know when to stop collecting input as heex - just to be clear :)  

Sorry typing from my phone but I hope that makes sense. Btw people started using MDEx with the phoenix_heex extension enabled so I'll push a fix here and there as reports come in for stuff I didn't anticipate.

kivikakk:
Understood, of course! (And always happy to be led by you on these changes :))

Btw, I just noticed [mdelixir.dev](https://mdelixir.dev)! This is an amazingly good look 😻 Time to add a link to [comrak.ee](https://comrak.ee)!

One thing I'd mention (as former GFM maintainer and today the GLFM maintainer) is that you could call MDEx's support for GitLab Flavored Markdown just as good as its support for GitHub Flavored Markdown, i.e. —

image

— no reason why these both can't be `[x]` :) There are "technically" parts of GLFM that aren't captured in Comrak itself, but the same applies to GFM and cmark-gfm. You can't get any closer than what's in MDEx, and anything not based on it/Comrak to start with would have a different foundation to GLFM itself, so as the maintainer I say it's as good as it gets ;)

leandrocp:
Hi @kivikakk 

> Btw, I just noticed [mdelixir.dev](https://mdelixir.dev/)! This is an amazingly good look 😻 Time to add a link to [comrak.ee](https://comrak.ee/)!

Thanks!

> support for GitLab Flavored Markdown just as good as its support for GitHub Flavored Markdown

Interesting, that's good to know. Already updated [mdelixir.dev](https://mdelixir.dev) and I also noticed the "powered by ..." was too hidden so I moved it from the footer to the header.

PR #737, opened 2026-01-21
Don't wrap text in table cells

cinerea0:
Currently, formatting a file in-place while using the `--width` flag wraps text in table cells at that width, which breaks table rendering. E.g., running `comrak -i --gfm --width 80` on the following:

```md
| option | description |
| ------ | ----------- |
| --width | Wraps at the given column number |
| --gfm | Use GFM-style quirks in output HTML, such as not nesting <strong> tags, which otherwise breaks CommonMark compatibility |
```
will produce:
```md
| option | description |
| --- | --- |
| --width | Wraps at the given column number |
| --gfm | Use GFM-style quirks in output HTML, such as not nesting <strong>
tags, which otherwise breaks CommonMark compatibility |
```

This PR checks if a text node is in a table cell, and passes that check to the `wrap` parameter of `output`, which prevents line breaks from being added to table cells.

### DIFF
--- a/src/cm.rs
+++ b/src/cm.rs
@@ -440,6 +440,8 @@ impl<'a, 'o, 'c, 'w> CommonMarkFormatter<'a, 'o, 'c, 'w> {
         let next_is_block = node
             .next_sibling()
             .is_none_or(|next| next.data().value.block());
+        let text_in_cell = node_matches!(node, NodeValue::Text(..))
+            && parent_node.is_some_and(|n| node_matches!(n, NodeValue::TableCell));
 
         match node.data().value {
             NodeValue::Document => (),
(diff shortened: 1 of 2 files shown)


kivikakk:
Thank you, this looks great! Would you be able to add a spec, so we can ensure this doesn't regress?

cinerea0:
Sorry, would that mean adding a new test, or something else?

kivikakk:
Yes, it would! A new test at the bottom of `src/tests/commonmark.rs` which turns on the table extension and sets the width parameter, and ensures that the wrapping behaviour doesn't apply to table lines that exceed the width.

kivikakk:
Thanks so much, this is great! :heart:

PR #739, opened 2026-01-26
Add "css" to CLI's syntax highlighting options; set CSS as the default highlighting mode

gjtorikian:
Couple of changes here, the title says them all.

I changed the default syntax highlighting behavior to use CSS classes, which is a breaking change, but I think it makes sense. In addition to browser rendering performance (CSS classes are less disruptive to apply than inline styles), every `write_pre_tag` call has to go look up the theme information to get the background color.


kivikakk:
Thank you, all SGTM! 🤍

PR #743, opened 2026-02-04
support raw nodes with children

JeanMertz:
This allows creating an AST within which `Raw` nodes can have children. This allows e.g. to wrap existing nodes into `Raw` nodes (in my case, I use this when printing markdown to a terminal to inject escape characters).


kivikakk:
Sounds good to me :)

PR #751, opened 2026-02-19
Support language-specific codefence renderers

neilberkman:
Supersedes #534.

This continues the original work from @tt and rebases it cleanly onto current `main`.

## What this PR adds
- `RenderPlugins::codefence_renderers` to register language-specific codefence renderers.
- `CodefenceRendererAdapter` integration in HTML fenced-code rendering.
- `CodefenceRendererAdapter::write` receives parsed `lang` + `meta` plus code and optional sourcepos.

## Rendering precedence
- built-in `math` handling
- language-specific renderer (when `lang` is non-empty and registered)
- syntax highlighter adapter
- default `<pre><code>` rendering

## Included follow-ups
- conflict resolution against current `main`
- expanded tests for precedence/sourcepos/fallback/empty-info behavior
- changelog entry under `v0.51.0 - Unreleased`
- `cargo fmt` clean and `cargo clippy -- -Dwarnings` clean

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -495,21 +495,18 @@ fn render_code_block<T>(
             let literal = &ncb.literal;
 
             if !info.is_empty() {
-                let lang_str = lang;
-                let info_str = meta;
-
                 if context.options.render.github_pre_lang {
-                    pre_attributes.insert("lang", lang_str.into());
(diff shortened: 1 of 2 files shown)


kivikakk:
Thank you for taking this over the line, this looks really good!

PR #752, opened 2026-02-19
Support comma-delimited language tokens in syntect

neilberkman:
Follow-up to #328, addressing #246.

Background:
- `cmark-gfm` uses whitespace-delimited info parsing and preserves `language-rust,ignore`.
- GitHub rendering still highlights `rust,ignore` as Rust.

This PR keeps Comrak's HTML class behavior unchanged while improving syntect syntax selection by splitting the language token on the first comma for lookup only.

What changed:
- `SyntectAdapter::write_highlighted`: use the part before `,` when selecting syntax.
- Add a regression test for fenced code using `rust,ignore` to ensure Rust highlighting applies.

Notes:
- This is the plugin-scoped approach from #328's discussion (no global parsing behavior change).
- Includes co-author attribution to the original contributor.

### DIFF
--- a/src/plugins/syntect.rs
+++ b/src/plugins/syntect.rs
@@ -86,7 +86,7 @@ impl SyntaxHighlighterAdapter for SyntectAdapter {
         let fallback_syntax = "Plain Text";
 
         let lang: &str = match lang {
-            Some(l) if !l.is_empty() => l,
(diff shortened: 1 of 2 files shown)


kivikakk:
Simple. Thanks!

PR #753, opened 2026-02-21
Fix off-by-one in hex entity digit limit

neilberkman:
## Summary

- Hex character references (`&#xHHHH;`) must have 1-6 hex digits per [CommonMark spec section 2.5](https://spec.commonmark.org/0.31.2/#entity-and-numeric-character-references), but the decimal branch's `(1..=7)` check was unconditionally OR'd, so 7-digit hex entities were also accepted
- Extract `is_hex` and gate each branch: hex allows 1-6 digits, decimal allows 1-7
- `&#x0000041;` (7 hex digits) was incorrectly parsed as `A`; now correctly treated as literal text

### DIFF
--- a/src/entity.rs
+++ b/src/entity.rs
@@ -34,10 +34,12 @@ pub fn unescape(text: &str) -> Option<(Cow<'static, str>, usize)> {
             0
         };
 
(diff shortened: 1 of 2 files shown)


kivikakk:
Hah. This is a very funny bug to find. Thank you!

PR #757, opened 2026-02-25
feat: Replace `unicode_categories` with `finl_unicode` for Unicode character categories

Martin005:
This PR updates the Unicode character category handling by switching from the [`unicode_categories` crate](https://crates.io/crates/unicode_categories) to the [`finl_unicode` crate](https://crates.io/crates/finl_unicode), which provides the same functionality.

The `unicode_categories` crate:
- was last [updated over 9 years ago](https://crates.io/crates/unicode_categories/versions),
- [isn't publishing new updates on crates.io](https://github.com/swgillespie/unicode-categories/issues/12),
- uses old Unicode data (hard to say which version, but it is older than Unicode 14, as a new version with [that commit hasn't been published yet](https://github.com/swgillespie/unicode-categories/commits/master/)).

The `finl_unicode` crate, on the other hand:
- is [updated frequently](https://crates.io/crates/finl_unicode/versions) (5 months ago the last time)
- uses the [newest Unicode 17 in the used 1.4.0 version](https://github.com/dahosek/finl_unicode?tab=readme-ov-file#version-history)
- according to [its benchmarks](https://github.com/dahosek/finl_unicode?tab=readme-ov-file#unicode-categories), it is **at least 20x faster** than `unicode_categories` (even more on some types of text)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -323,6 +323,7 @@ dependencies = [
  "codspeed-divan-compat",
  "emojis",
  "entities",
+ "finl_unicode",
  "fmt2io",
  "glob",
  "jetscii",
@@ -339,7 +340,6 @@ dependencies = [
  "syntect",
  "toml",
  "typed-arena",
- "unicode_categories",
  "xdg",
 ]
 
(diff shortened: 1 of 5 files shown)


kivikakk:
Thanks, this looks really good! 🤍

PR #763, opened 2026-03-07
Add Typst formatter

neilberkman:
This adds a Typst source formatter to comrak.

I realize this may read as bloat for core, so if you think this does not belong here, or would be better as a feature flag or separate crate, feel free to close. I am a big Typst user and wanted to put together a concrete implementation to see whether it earns its keep.

What this adds:
- `format_typst` / `format_typst_with_plugins`
- `markdown_to_typst` / `markdown_to_typst_with_plugins`
- CLI support via `-t typst`
- Typst-specific tests

Current coverage:
- headings, paragraphs, emphasis, strong, inline code, code blocks
- links and images
- ordered and unordered lists
- task lists
- block quotes and alerts
- footnotes and inline footnotes
- tables
- description lists
- wikilinks
- shortcodes
- heading IDs and internal label links
- raw `typst` fences passthrough
- best-effort translation of common LaTeX-style math in Markdown math

A few notes:
- This is only a Typst source formatter. It does not embed or depend on the Typst compiler.
- Math support is best-effort for the common Markdown / LaTeX-ish subset, not full TeX compatibility.
- Where there is no clean Typst analogue, the formatter either degrades conservatively or preserves raw output.

Why I think this is useful:
- Markdown is a common interchange format, including for LLM-generated content.
- Typst is a strong target for deterministic typeset output.
- This makes comrak usable in pipelines where Markdown is the source and Typst is the publishing/layout layer.

I also checked size impact because that was my first concern:
- release `comrak` `.rlib`: about `+271 KB`
- release `comrak` CLI binary: about `+55 KB`
- in a stripped downstream consumer binary that did not call any Typst API, the size delta was effectively negligible in my local test

If you would prefer, I am happy to rework this behind a feature flag or move it into a companion crate instead of keeping it in core.

Tests run:
- `cargo test`
- `cargo test typst`

### DIFF
--- a/src/tests/pathological.rs
+++ b/src/tests/pathological.rs
@@ -5,7 +5,7 @@ use ntest::timeout;
 
 // input: python3 -c 'n = 50000; print("*a_ " * n)'
 #[test]
-#[timeout(8000)]
+#[timeout(4000)]
 fn pathological_emphases() {
     let n = 50_000;
     let input = "*a_ ".repeat(n);


kivikakk:
Hiya, thank you so much for this!

> I realize this may read as bloat for core, so if you think this does not belong here, or would be better as a feature flag or separate crate, feel free to close. I am a big Typst user and wanted to put together a concrete implementation to see whether it earns its keep.

I myself am a big fan of Typst (though I don't get as much opportunity to use it as I'd like); I think it could do with more exposure! So I'm happy to include this in core.

It'll take me a couple days to get time to review this with the care it deserves; will let you know. In the mean time, thanks again!

PR #764, opened 2026-03-09
docs: Explain column counting in LineColumn and fix shortcode context link

Martin005:
This PR clarifies documentation for `column` position handling in the `LineColumn` struct and makes the link in `NodeShortCode` documentation clickable.


kivikakk:
Thanks so much, @Martin005! 🐰

PR #769, opened 2026-03-10
feat: Add compact_html render option to suppress pretty-printing newlines

### LINKED ISSUE #768: There is no way to disable pretty HTML output
By default Comrak `comrak::format_html()` will output HTML formatted with each element opening and closing tags in a new line. Is there any way to ouput the full HTML output in a single line?

xvchris:
Adds a `compact_html` boolean to `Render` that turns off the `\n` characters comrak normally writes after block-level closing tags (`</p>`, `</li>`, etc).

When the option is off (default) nothing changes. When it's on you get output like:

```html
<h1>Hello</h1><p>World.</p>
```

instead of:

```html
<h1>Hello</h1>
<p>World.</p>
```

Exposed as `--compact` on the CLI.

Closes #768

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -427,10 +427,12 @@ fn render_block_quote<T>(
         context.cr()?;
         context.write_str("<blockquote")?;
         render_sourcepos(context, node)?;
-        context.write_str(">\n")?;
+        context.write_str(">")?;
+        context.lf()?;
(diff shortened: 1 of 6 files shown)


kivikakk:
Thanks so much! Looks perfect to me :)

PR #773, opened 2026-03-14
feat: add MathAdapter plugin and pulldown-latex MathML renderer

### LINKED ISSUE #537: Allow rendering mathematics as mathml in html
This is probably a duplicate of #208, but I'm not quite sure why that one is closed.

I think it'd be nice to bundle a mathml renderer since this seems like a common use-case, and it'd be easier to use from downstream libraries wrapping this one.

vimzh:
## Summary

Adds a `MathAdapter` plugin trait and a built-in `PulldownLatexAdapter` that converts LaTeX math to MathML, addressing #537.

- **`MathAdapter` trait** in `adapters.rs` — custom rendering hook for all math content (`$...$`, `$$...$$`, `` $`...`$ ``, ` ```math `)
- **`math_renderer` field** in `RenderPlugins` — follows the existing `SyntaxHighlighterAdapter` + `SyntectAdapter` pattern
- **`PulldownLatexAdapter`** behind optional `mathml` cargo feature — uses [`pulldown-latex`](https://crates.io/crates/pulldown-latex) (actively maintained, MathML Core spec compliant)
- Graceful fallback: invalid LaTeX renders as escaped text in `<span class="math-error">` instead of failing
- LaTeX source preserved in `<annotation encoding="application/x-tex">` by default (configurable)
- Zero impact when adapter is not set — existing behavior is completely unchanged

## Usage

```rust
use comrak::plugins::pulldown_latex::PulldownLatexAdapter;
use comrak::{Options, markdown_to_html_with_plugins, options};

let adapter = PulldownLatexAdapter::default();
let mut options = Options::default();
options.extension.math_dollars = true;
let mut plugins = options::Plugins::default();
plugins.render.math_renderer = Some(&adapter);

let html = markdown_to_html_with_plugins("$E=mc^2$", &options, &plugins);
// <p><math display="inline"><semantics><mrow>...</mrow>
//   <annotation encoding="application/x-tex">E=mc^2</annotation>
// </semantics></math></p>
```

## Design decisions

- **Plugin adapter (not render option)**: Follows the `SyntaxHighlighterAdapter` pattern. Users can implement `MathAdapter` for custom renderers (KaTeX, MathJax, etc.) without depending on pulldown-latex.
- **`pulldown-latex` over `latex2mathml`**: latex2mathml hasn't been updated since 2020. pulldown-latex is actively maintained (last release Nov 2024), follows MathML Core spec, and has a clean pull-parser API.
- **Optional feature**: `mathml` feature flag keeps the dependency opt-in, not in `default`.

## Test plan

- [x] 6 mock adapter tests (inline dollar, display dollar, code inline, code block, sourcepos, fallback)
- [x] 5 pulldown-latex integration tests (inline, display, code block, no-annotation, sourcepos)
- [x] 1 doctest on `PulldownLatexAdapter`
- [x] All 606 existing tests pass without `mathml` feature (zero regression)
- [x] All 611 tests pass with `mathml` feature
- [x] Clippy clean, `cargo fmt` clean

Closes #537

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -332,6 +332,7 @@ dependencies = [
  "phf",
  "phf_codegen",
  "pretty_assertions",
+ "pulldown-latex",
  "rustc-hash",
  "shell-words",
  "slug",
@@ -836,6 +837,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "535d180e0ecab6268a3e718bb9fd44db66bbbc256257165fc699dadf70d16fe7"
 dependencies = [
  "unicode-ident",
+]
+
+[[package]]
+name = "pulldown-latex"
+version = "0.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3b8bc0583825170e3f560701d966dc2f0e3a16946da371e37d30d3ad7207fb7e"
+dependencies = [
+ "bumpalo",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,7 @@ caseless = "0.2"
 fmt2io = { version = "1.0.0", optional = true }
 jetscii = "0.5.3"
 phf = "0.13"
+pulldown-latex = { version = "0.7", optional = true, default-features = false }
 rustc-hash = "2"
 smallvec = "1.13"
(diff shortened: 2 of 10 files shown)


kivikakk:
Please note the contributing guidelines.

PR #779, opened 2026-03-24
feat: Add parse option for char-based columns in Sourcepos

### LINKED ISSUE #777: (Proposal) Parse option to make sourcepos column char-based
As discussed in https://github.com/kivikakk/comrak/discussions/762, `comrak` uses amount of UTF-8 bytes for the value in [`LineColumn::column`](https://docs.rs/comrak/latest/comrak/nodes/struct.LineColumn.html#structfield.column).

But if you want to use `comrak` to pinpoint the exact location in the source file, the user would expect the `column` to point at the readable character, not at the position of a UTF-8 byte.

When you have a Markdown file that contains only the `好` character (U+597D - `0xE5 0xA5 0xBD` in UTF-8 bytes), `comrak` puts `1:1-1:3` as the sourcepos.

Contrary to that, if you open the same Markdown file in a text editor (e.g. Notepad++), the end column is at 2, and the position is 4. `comrak` treats the range as end-inclusive, so it puts `1:1-1:3` instead of `1:1-1:4` as the sourcepos.

Image
<br/>
<br/>

My proposal is to **create a parse option** (disabled by default) which, when enabled, would make the column value in `LineColumn` **character-based**, not based on UTF-8 bytes.

That would make the Markdown example above have sourcepose `1:1-1:1` (as the end-inclusivity would be maintained).

Martin005:
This PR introduces a new parsing option to report source position columns as Unicode character counts instead of UTF-8 byte offsets.

I created a lot of tests for different node values / extensions, but the tests are not exhaustive. Please let me know if I should cover more things 🙂 

Closes #777

### DIFF
--- a/fuzz/fuzz_targets/all_options.rs
+++ b/fuzz/fuzz_targets/all_options.rs
@@ -134,6 +134,7 @@ struct FuzzParseOptions {
     default_info_string: bool,
     broken_link_callback: bool,
     escaped_char_spans: bool,
+    sourcepos_chars: bool,
 }
 
 impl FuzzParseOptions {
@@ -159,6 +160,7 @@ impl FuzzParseOptions {
                 None
             },
             escaped_char_spans: self.escaped_char_spans,
+            sourcepos_chars: self.sourcepos_chars,
         }
     }
 }


kivikakk:
> I created a lot of tests for different node values / extensions, but the tests are not exhaustive. Please let me know if I should cover more things 🙂

Your tests are always super exhaustive to my eyes; I really appreciate your efforts!

Martin005:
> > I created a lot of tests for different node values / extensions, but the tests are not exhaustive. Please let me know if I should cover more things 🙂
> 
> Your tests are always super exhaustive to my eyes; I really appreciate your efforts!

By not being exhaustive, I meant that the test cases don't test every single node value / extension. And there isn't a case for a long Markdown document with multiple node values.
But I don't think there should be a situation where the `sourcepos_chars` gives an incorrect value as the implementation is very straightforward (famous last words 😀)

Issue #780, opened 2026-03-26
Add fenced container extension


P-SiZK:
Would it be possible to add a fenced container extension?
This would be a useful syntax for custom extensions, as discussed in the [Generic directives/plugins syntax](https://talk.commonmark.org/t/generic-directives-plugins-syntax/444) proposal.
The simplest form mirrors fenced code blocks, using colons instead of backticks:

```md
:::foo bar
Lorem ipsum
:::
```

The HTML output below is one possible rendering.

```html
<div class="foo bar">
<p>Lorem ipsum</p>
</div>
```

This syntax is already used in tools like [Docusaurus](https://docusaurus.io/docs/markdown-features/admonitions) and [VuePress](https://v1.vuepress.vuejs.org/guide/markdown.html#custom-containers).
If this seems like a reasonable addition, I'd be happy to open a pull request.

kivikakk:
Totally reasonable! Please do open a PR 😸

digitalmoksha:
@P-SiZK I think adding support for directives is a great idea. But I have a couple thoughts:

1. I think we should follow the more common syntax for them, such as in [Docusarus](https://docusaurus.io/docs/markdown-features/admonitions) or [micromark](https://github.com/micromark/micromark-extension-directive?tab=readme-ov-file#directive)

   ```
   ::: name [inline-content] {key=val}
   contents, which are sometimes further block elements
   :::
   ```

   For example in Docusarus, you can create a `note` admonition using

   ```
   :::warning

   Hello world

   :::
   ```

   And you can change the title text of the admonition by specifying the `inline-content`

   ```
   :::warning[This is a Warning]

   Hello world

   :::
   ```

   See also the [original syntax proposal](https://talk.commonmark.org/t/generic-directives-plugins-syntax/444).  Note that [commonmark-hs](https://github.com/jgm/commonmark-hs/blob/master/commonmark-extensions/test/fenced_divs.md) supports a `fenced_divs` option (which just allows specifying a `div` with attributes), as does [pandoc](https://pandoc.org/MANUAL.html#extension-fenced_divs).  In fact there is an [issue](https://github.com/jgm/pandoc/issues/7480) for extending pandoc to use the above syntax.

[...]

PR #782, opened 2026-03-29
Add block directive extension

P-SiZK:
Closes https://github.com/kivikakk/comrak/issues/780

This implementation referred to the existing fenced code block and multiline block quote implementations.


kivikakk:
Thank you! I've looked through this thoroughly — this is totally OK for merge.

I do think it's worth considering @digitalmoksha's ideas in https://github.com/kivikakk/comrak/issues/780#issuecomment-4150793187, but they don't need to block merging since (a) they should only be implemented if there's someone who wants or needs them and is willing to do it, and that might not be you!, and (b) it's fully possible to achieve the structure he's talked about there using this implementation and a custom renderer which further parses the `info` string.

One thing I will note is it's not possible to have `:` _anywhere_ in the info string; I wonder if we might want to loosen that up in future? e.g. permit `:` as long as there's a non-`:` separating it from the fence marker.

```diff
-    [:]{3,} / [^:\r\n\xff]*[\r\n\xff] { return Some(cursor); }
+    [:]{3,} / ([^:\r\n\xff][^\r\n\xff]*)? [\r\n\xff] { return Some(cursor); }
```

digitalmoksha:
I tend to fall on the side of using a generally accepted syntax from the get go, rather than possibly introducing backward compatibility issues or having to support an older style syntax forever. In the generally accepted syntax, I don’t think ‘:::foo bar’ would be valid. But I could be wrong. 

Either way, glad something has been added.

kivikakk:
I understand! I have always and continue to go by "optimistic merging" (per [Contributing](https://github.com/kivikakk/comrak?tab=readme-ov-file#contributing)), and really want to avoid adding workload to contributors that isn't actually responsive to their needs — especially a first-time contributor.

As long as Comrak is 0.x, I'm quite happy to introduce breaking changes, especially in very new functionality, so I'm not concerned about supporting things forever at this stage.

P-SiZK:
Thanks for the feedback!

I initially referenced VuePress, which turned out to use [markdown-it-container](https://github.com/markdown-it/markdown-it-container) under the hood, so I ended up following that simpler approach where everything after `:::` is treated as a single info string.
As you mentioned, a custom renderer can handle the directives it supports, so I thought that was flexible enough without parsing `name` and `[inline-content]` separately.

If `[inline-content]` support were added, I'm curious how you think compatibility with the markdown-it-container style would work — since that allows an info string to contain multiple words, like `:::foo bar`.
Do you think `name` should be strictly a single token, which would make `:::foo bar` invalid?

Please let me know if I'm misunderstanding the generally accepted syntax you're referring to.

PR #791, opened 2026-04-06
Escape '~', '~~~' and ':::' markdown patterns

sandmor:
This PR address the same issue than #788, rethinking the approach based on the comments from @kivikakk. In this case we limit ourselves to add '~' to the characters that need to be escape, and ':' in case of being at the beginning of a block and the block directive extension being enabled.

Additionally '~' is escaped inline to avoid issues with ~strikethrough~ text in markdown.

Before these changes:
```sh
$ printf '\~~~' | comrak
<p>~~~</p>
$ printf '\~~~' | comrak -t commonmark
~~~
$ printf '\~~~' | comrak -t commonmark | comrak
<pre style="background-color:#2b303b;"><code></code></pre>
```
And after:
```sh
$ printf '\~~~' | comrak
<p>~~~</p>
$ printf '\~~~' | comrak -t commonmark
\~\~\~
$ printf '\~~~' | comrak -t commonmark | comrak
<p>~~~</p>
```

### DIFF
--- a/src/cm.rs
+++ b/src/cm.rs
@@ -300,10 +300,14 @@ impl<'a, 'o, 'c, 'w> CommonMarkFormatter<'a, 'o, 'c, 'w> {
                     || c == '>'
                     || c == '\\'
                     || c == '`'
+                    || c == '~'
                     || c == '!'
                     || (self.options.extension.autolink && c == '@')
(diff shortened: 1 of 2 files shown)


kivikakk:
Awesome, thank you so much! 🤍🐰

PR #792, opened 2026-04-10
Make `--inplace` accept multiple arguments

ilkecan:
`--inplace` not accepting multiple files was causing problems with [prek](https://github.com/j178/prek/issues/1471) and I saw that this was already considered as a future goal in https://github.com/kivikakk/comrak/pull/377#issuecomment-2106296750.

---

Changes:
- update `--inplace` help text from singular to plural
- relax validation so `--inplace` only rejects stdin / no input files
- process each `--inplace` file independently in CLI argument order

This keeps non-`--inplace` behavior unchanged. Multiple files passed without `--inplace` are still concatenated and parsed as a single document.

`comrak --inplace a.md b.md` now behaves like running the existing single-file `--inplace` flow once for `a.md` and once for `b.md`. If a later file fails, earlier rewritten files are not rolled back.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ Options:
           [default: /home/runner/.config/comrak/config]
 
   -i, --inplace
-          Reformat a CommonMark file in-place
+          Reformat CommonMark files in-place
 
       --hardbreaks
           Treat newlines as hard line breaks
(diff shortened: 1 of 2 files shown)


kivikakk:
Thanks, appreciate this!

PR #803, opened 2026-05-15
Improve heading anchor accessibility

### LINKED ISSUE #573: Heading anchors should include the inert property
Screen readers can't handle the anchors as they are as they retain focus when you go to them, the output html should include the [`inert`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/inert) property to prevent any interactibility since they're not a visible item in the page but a functional one in the hierarchy.

MukundaKatta:
## Summary
- move generated heading anchor links after the heading content
- replace aria-hidden empty anchors with accessible labels and data-heading-content
- update header ID docs and tests for the new output

Fixes #573

## Testing
- cargo fmt --check
- cargo test header_id_prefix
- cargo test html::
- git diff --check

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -609,7 +609,7 @@ fn render_heading<T>(
                 write!(context, "<h{}", nh.level)?;
                 render_sourcepos(context, node)?;
                 context.write_str(">")?;
-
+            } else {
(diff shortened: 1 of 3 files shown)


kivikakk:
Please review the contributing guidelines.

PR #817, opened 2026-07-06
Deduplicate attribute value parsing

KevSlashNull:
This deduplicates the attribute value parsing by extracting the `match kind` block to a new helper function, which reduces code reuse as complained about in the removed comment.

### DIFF
--- a/src/parser/attributes.rs
+++ b/src/parser/attributes.rs
@@ -2,6 +2,14 @@ use std::mem;
 
 use crate::nodes::Attributes;
 
+#[derive(Default)]
+enum Kind {
+    #[default]
+    Id,
+    Class,
+    Pair(String),
+}
+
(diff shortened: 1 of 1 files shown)


src/parser/attributes.rs
```diff
@@ -142,6 +133,14 @@ pub fn parse(input: &str) -> Option<(Attributes, usize)> {
     }
 }
 
+fn parse_attribute_value(kind: &mut Kind, value: &mut String, attrs: &mut Attributes) {
```
KevSlashNull, line 136: on second though, `parse_value` might've been better but that feels too ambiguous. what even is a value (ignoring that `State::Value` enum constant at the top :p)

kivikakk: :D Either is fine by me!

kivikakk (approved):
Thanks so much! 🐰

PR #822, opened 2026-07-11
feat: add option to emit semantic HTML for alerts

### LINKED ISSUE #821: Match `docutils`/`rST` HTML output for admonitions/alerts
Admonitions are an old reStructuredText feature. GFM’s “alerts” are very clearly directly copied from it: the types it uses (note, tip, caution, …) are a subset of rST’s. If it does, CommonMark will likely adopt them under the older existing name, “admonitions”.

Using the main `rST` implementation `docutils` to create HTML results in this:

```html
<aside class="admonition note">
<p class="admonition-title">Note</p>
<p>The body.</p>
</aside>
```

It would be nice if people could use comrak to generate HTML that seamlessly integrates into existing style sheets. If that’s wanted, we could add and option to emit them in this existing format (which also happens to be more semantic, using `aside`s instead of `div`s). As said: supported types are an exact subset of rST’s, so integration will be seamless.

flying-sheep:
fixes #821.

Design question: Maybe it should be made more generic, e.g. rename it to `HtmlStyleType` and have it affect `tasklist_classes` too – for some reason you don’t add a `markdown-` prefix for that class, only for alerts. But then it would be a breaking change as its default would either change `render.tasklist_classes` or `extension.alerts`.


kivikakk:
> Design question: Maybe it should be made more generic, e.g. rename it to `HtmlStyleType` and have it affect `tasklist_classes` too – for some reason you don’t add a `markdown-` prefix for that class, only for alerts. But then it would be a breaking change as its default would either change `render.tasklist_classes` or `extension.alerts`.

Maybe! The naming of the tasklist classes there was pretty arbitrary and decided upon by whoever implemented them at the time; I kind of think Comrak itself should probably not have much in the way of opinions about classes _at all_, and anyone who really wants them probably wants to implement a custom formatter.

But for now, we have what we have. (I'm pretty OK with breaking changes FWIW! What's the point of pre-1.0 if we can't have fun?)

Thanks so much for this PR! I really appreciate it. I'll add a CLI entrypoint and merge :)

kivikakk:
This is part of Comrak [v0.54.0](https://github.com/kivikakk/comrak/releases/tag/v0.54.0)!

flying-sheep:
Thanks! This will help tools that support different markup languages!

PR #828, opened 2026-08-18
Support Obsidian wikilink embeds (`![[target]]`)

### LINKED ISSUE #535: Obsidian wikilink images ![[image.jpg]]
Hi @kivikakk, thanks so much for your work here.

Image nodes are emitted for the normal `` image tags, but not for `![]` wikilink image tags.

✅  Markdown links 
```md
![image:width:20%](fido.jpg)
```

❌  wikilinks 
```md
![[fido.jpg|image:width:20%]]
```

I wonder if it might be a simple change to support this? Thank you!

ChrisJr404:
Closes #535.

Wikilinks written with a leading `!` (Obsidian's embed syntax, `![[target]]`) previously weren't recognised: the `!` was consumed as an ordinary image marker, so no wikilink node was produced. This adds support for parsing them as embeds when a wikilinks extension is enabled.

Following the discussion on #535, the parsing surface is the important part — `NodeWikiLink` now carries an `embed: bool` so downstream consumers can tell embeds apart and decide how to handle the target (image, transcluded note, canvas, etc.). Since the issue is specifically about image embeds and you pointed at `image_url_rewriter`, embeds also render as images out of the box.

### What's included

- **Parsing:** `![[target]]` is parsed as a wikilink with `embed = true` (both `wikilinks_title_after_pipe` and `wikilinks_title_before_pipe` modes). Plain `[[target]]` keeps `embed = false`. The source position includes the leading `!`. If the `![[…` doesn't form a valid wikilink, it falls back to the previous image-marker behaviour.
- **`NodeWikiLink::embed`:** new field, defaults to `false`.
- **HTML:** embeds render as ``, with the label as alt text and `image_url_rewriter` applied to the URL (mirroring `render_image`). Dangerous URLs are sanitised just like images. Non-embed wikilinks are byte-for-byte unchanged.
- **XML:** embeds get an `embed="true"` attribute.
- **CommonMark round-trip:** embeds are re-emitted with the leading `!`.
- **Docs:** the `wikilinks_title_after_pipe` option gains a note + doctest for embeds.

### Tests

Added to `src/tests/wikilinks.rs`: HTML rendering for both pipe modes, `image_url_rewriter` interaction, URL sanitisation, the XML `embed` attribute, source positions, and regressions confirming that `[[target]]` still renders as a link and that `![[…]]` is inert without the extension. Full suite (`cargo test`) and `cargo fmt --check` pass; no new clippy warnings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,17 @@ Behind the scenes:
 
 -->
 
+# [Unreleased]
+
+Parser changes:
+
+* Wikilink embeds (`![[target]]`) are now recognised when a wikilinks extension is enabled, matching Obsidian's embed syntax.
+
+Changed APIs:
+
+* `NodeWikiLink` gained an `embed` field indicating whether the wikilink was written as an embed (`![[target]]`). Embeds are rendered as images in HTML output (respecting `image_url_rewriter`) and round-trip through the CommonMark and XML formatters.
+
+
 # [v0.54.0] - 2026-07-12
 
 Changed APIs:
--- a/src/cm.rs
+++ b/src/cm.rs
@@ -1013,6 +1013,9 @@ impl<'a, 'o, 'c, 'w> CommonMarkFormatter<'a, 'o, 'c, 'w> {
 
     fn format_wikilink(&mut self, nl: &NodeWikiLink, entering: bool) -> fmt::Result {
         if entering {
+            if nl.embed {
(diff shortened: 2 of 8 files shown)


kivikakk:
<https://github.com/kivikakk/comrak/blob/main/CONTRIBUTING.md#policy-on-llm-contributions>.

Issue #832, opened 2026-09-07
Autolink: three differences from cmark-gfm


akiomik:
Comparing comrak 0.54.0 against github/cmark-gfm master turned up three cases where the two disagree.
Filing them together because they came out of one comparison; happy to split.

All outputs below are from comrak 0.54.0 with `extension.autolink = true` and everything else default.

---

## 1. `url_match` no longer allows a domain without a dot

```
see http://localhost/x now
  -> <p>see http://localhost/x now</p>
```

GitHub links this.
So did comrak up to 0.43.0.

cmark-gfm asks a dot of a `www.` host and not of a scheme'd one.
The two call sites pass different constants (`extensions/autolink.c`):

```c
link_end = check_domain(data, size, 0);                          /* www_match */
domain_len = check_domain(data + link_end, size - link_end, 1);  /* url_match */
```

`check_domain` says why:

```c
  if (allow_short) {
    /* We don't need a valid domain in the strict sense (with
     * least one dot; so just make sure it's composed of valid
     * domain characters and return the length of the the valid
     * sequence. */
    return i;
  } else {
    /* a valid domain needs to have at least a dot.
     * that's as far as we get */
    return np ? i : 0;
  }
```

comrak passes `relaxed_autolinks` at both call sites (`src/parser/autolink.rs`):

```rust
let mut link_end = check_domain(&subject.input[i + 4..], relaxed_autolinks)? + 4;  // www_match
let mut link_end = check_domain(&subject.input[i + 3..], relaxed_autolinks)? + 3;  // url_match
```

`relaxed_autolinks` is false by default, so `url_match` asks for a dot.

This is a change, not a long-standing difference:

| version | www_match | url_match |
| --- | --- | --- |
| 0.11.0 - 0.43.0 | `false` | `true` |
| 0.44.0 - 0.54.0 | `relaxed_autolinks` | `relaxed_autolinks` |

It changed in #618.
That PR's description says:

> This second commit covers an issue where the number of periods in a domain was
> not validated for an extended URL autolink. This was clearly intentional, as
> there was a dedicated parameter to loosen the check, and tests for it. However,
> it means that bare schemes such as "http://" are autolinked. I can't see any
> reference to this modified validation in the spec.

The spec text does say a valid domain needs a dot.
cmark-gfm does not apply that to `url_match`.
#190 covered the same ground in 2021 and was closed by #191:

> This is why GitHub itself doesn't support it either; examples:
[...]

PR #833, opened 2026-09-08
autolink: three cmark-gfm parity fixes

YuEfSaEDU:
With `extension.autolink` enabled, comrak diverged from cmark-gfm in three cases (#832).

`url_match` now calls `check_domain` with `allow_short = 1` as cmark-gfm does, so dot-less hosts such as `http://localhost/x` autolink again (they did through 0.43.0; the requirement of a dot dates to #618), while a new check mirroring the host-character test in `sd_autolink_issafe` keeps bare `http://` from linking; `http://user:pass@host/` is likewise matched whole instead of falling through to the email matcher.

The scheme is now compared case-insensitively, matching `sd_autolink_issafe`, so `HTTP://www.example.com/` autolinks while `www.` matching stays case-sensitive.

Finally, the autolink matchers consult the bracket stack rather than a flag cleared by any `]`, mirroring `cmark_inline_parser_in_bracket`, so an inner `]` no longer lets an autolink run through an enclosing link or produce an `<a>` nested inside an `<a>`. Fixture tests are added for each case.

### DIFF
--- a/src/parser/autolink.rs
+++ b/src/parser/autolink.rs
@@ -427,13 +427,29 @@ pub fn url_match<'a>(
 
     if !relaxed_autolinks {
         let scheme = &subject.input[i - rewind..i];
-        let cond = |s: &&str| size - i + rewind >= s.len() && &scheme == s;
+        // As in cmark-gfm's `sd_autolink_issafe`, the scheme is compared
+        // case-insensitively.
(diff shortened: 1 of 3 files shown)


kivikakk:
This user does not appear to be submitting human-authored patches; closing per CONTRIBUTING.

IMG_1534

PR #836, opened 2026-09-17
Measure fenced block offset in columns, not bytes

szdytom:
`fence_offset` is consumed one column at a time by the block prefix parsers (`parse_code_block_prefix`, `parse_multiline_block_quote_prefix`, `parse_block_directive_prefix`), but it was recorded as the byte distance `first_nonspace - offset`. When a container prefix consumes only part of a tab, the byte count under-counts the indentation, so `add_line` materialises the rest of the tab as leading spaces in the block's content.

## Minimal reproduction

The indentation below is a single tab (`→`):

```markdown
- a

→```
→x
→```
```

commonmark.js (the dingus) produces:

```html
<ul>
<li>
<p>a</p>
<pre><code>x
</code></pre>
</li>
</ul>
```

Comrak produces `<pre><code> x`, with a leading space before `x`. The list marker eats two of the tab's four columns, so the fence is indented two columns; measuring in bytes gives one.

The same happens inside a block quote:

```markdown
> →```
> →x
> →```
```

commonmark.js renders `<pre><code>x`; Comrak renders `<pre><code> x`.

## Reference implementation

commonmark.js, which this parser is ported from, records the offset in columns:

```js
// lib/blocks.js
container._fenceOffset = parser.indent;
```

(`cmark` 0.31.2 happens to count bytes here as well, so this is one of the commonmark.js ↔ cmark divergences; the fix aligns Comrak with the dingus.)

## Fix

Use `self.indent` at all four `fence_offset` assignments: code fences, alerts, multiline block quotes, and block directives.

Tests added in `src/tests/regressions.rs` for the list and block quote reproductions.


kivikakk:
Thanks @szdytom! LGTM :)

PR #837, opened 2026-09-17
replace jetscii with memchr-n

Dr-Emann:
jetscii seems pretty much unmaintained, and doesn't accelerate aarch64


src/parser/mod.rs
```diff
@@ -198,7 +201,7 @@ where
         self.total_size = end;
 
         let mut ix = 0;
-        let matcher = jetscii::bytes!(b'\r', b'\n');
+        let matcher = MATCHER.get_or_init(|| MemchrN::new(b"\r\n"));
 
         while ix < end {
             let mut eol = match matcher.find(&sb[ix..]) {
```
Dr-Emann, line 207: This could probably be rewritten to use an iterator rather than repeated `find` calls.

kivikakk:
Thanks for this PR, @Dr-Emann!

I have some hesitancy in accepting a very new dependency, but I read through everything and everything looks good! It'd be lovely to get acceleration on aarch64 too. I measure an 18% improvement vs jetscii on an M4 doing `parse_document + format_html` on the progit corpus!

There's some failures in CI (`tests::core::entities` etc. stemming from `html` emitting too much), but more importantly than that: `memchr-n` (and `fearless_simd`) declare `rust-version = "1.89"`, ahead of Comrak's MSRV, 1.85. We moved to that in January (https://github.com/kivikakk/comrak/pull/732), and I'm a _little_ hesitant to bump it again quite so soon. I say _little_ because 1.89 turned 1 year old in August, and I think I'm actually OK with that cadence 👍

PR #838, opened 2026-09-19
Update ranges for is_cjk to Unicode 18 & add fast path

tats-u:
https://github.com/tats-u/markdown-cjk-friendly/commit/caa37236186e0d294cd6da255a1bdde81dbc0f50

- Update ranges for CJK friendly emphasis to Unicode 18.
- Add a fast path for the most major scripts (e.g. European, Western Asian, South Asian) whose code points are less than those of NFD Korean (U+1100–U+11FF). I hope it works well.


kivikakk:
Thanks @tats-u! :)

